PR #5, opened 2018-09-14
Add conda recipe

pavlin-policar:
Make the package distributable with conda.

`pynndescent` is not installable with conda, which means that either I can could do that myself (which is a lot of maintaining) or I could copy the code here and distributed it with this package. While not ideal, the second option seems better.


pavlin-policar:
Waiting on https://github.com/conda-forge/fasttsne-feedstock to be set up.

PR #6, opened 2018-09-15
Travis CI: flake8 finds syntax errors and undefined names

cclauss:
Add [flake8](http://flake8.pycqa.org) tests to find Python syntax errors and undefined names.

__E901,E999,F821,F822,F823__ are the "_showstopper_" flake8 issues that can halt the runtime with a SyntaxError, NameError, etc. Most other flake8 issues are merely "style violations" -- useful for readability but they do not effect runtime safety.
* F821: undefined name `name`
* F822: undefined name `name` in `__all__`
* F823: local variable name referenced before assignment
* E901: SyntaxError or IndentationError
* E999: SyntaxError -- failed to compile a file into an Abstract Syntax Tree


pavlin-policar:
Thanks for your PR! I am not familiar with flake8, but it looks like a substitute for `pylint`?

If I understand correctly, this will stop the Travis build early if there are syntax errors in the code (is this typically an issue?) and report PEP8 violations as warnings in the Travis log? I am hesitant to add too many code quality tools since it's all too easy to set up many tools and end up doing nothing but tooling. Codacy runs pylint so we already have a linter in place.

I am struggling to see the added benefit of this. The real improvement might be stopping builds early, and it would make sense to add this to Appveyor as well since in my experience Travis builds are much faster than Appveyor builds but even still, most the time is spent setting up the environment so this wouldn't help out there. I don't think long build runtimes have ever really been a real issue though. Please help me see why this makes sense.

As for the benchmarks directory, I am considering removing this from the repository entirely. Most of the files in there are very, very messy and their only use is to load data sets (if you have them locally), run benchmarks and to experiment with the code. It is not included anywhere in the actual tSNE library itself, so this directory should be excluded from tests/linting or anything of that nature at the very least, if not removed entirely.

cclauss:
Yes.  The idea is to _fail fast_.  If a PR contains a syntax error or an undefined name then we want the contributor to get feedback fast so they can fix it fast.  No need to run longer tests, just fail the build quickly.  I consider both Circle CI and Travis to be faster at startup than the rather lethargic AppVeyor to I added the tests to Travis.  We do not need to run these tests in multiple CI systems as they would merely return identical results.  It is however useful to run the tests on both Python 2 and 3 if both are going to be supported as they each have different syntax and therefor different Abstract Syntax Trees.

A linter in a dynamic language should flag issues that the compiler would flag in a compiled language.  It should raise an alert if there is a syntax error or if an import is missing.  It should warn you that __basestring__, __cmp()__, __long__, __reload()__, __unicode__, __xrange()__ is not builtins in Python 3.  Codacy does none of these things that a compiler would.  It merely says that __async def func():__ is not valid Python.  It seems to be more legacy than the PyLint that it is based on.  In the hands of a strong practitioner, PyLint can catch many of the issues above but it take configuration and usually a lot of __# pylint: disable=xyz__ directives in the code.
The second round of flake8 tests is purely optional and that is why I __exit-zero__ it.  The vast majority of those issues can be automatically corrected with Black, AutoPEP8, or yapf.  The remaining ones are actually worth fixing but are not vital to runtime safety.

cclauss:
Removed the second round (exit-zero) flake8 tests.  We are down to just five "showstopper" tests discussed in the commit message above.  If Codacy can be configured to flag these issues, I would love to learn how to do that.

cclauss:
AppVeyor does Windows and Linux.  Travis does Mac and Linux or anything in a Docker container.  Circle focuses on making the Docker approach easier.

I like the Python 3.5+ approach — tuned to the times.

79 characters is for punch cards!  Black likes 88 which is sane.  My max-line-length is 127 so code can be read on GitHub without right-left scrolling.

I don’t use Codacy but it does have a following so I might be missing something.

pavlin-policar:
> Black likes 88 which is sane.

I was not aware of that, I may have to look into it then.

Thanks for the PR and for introducing me to flake8. If it turns out to be really good, I may end up replacing Codacy with it entirely. Thanks for this informative conversation!

Issue #10, opened 2018-09-17
Normalize PCA initialization to have small variance


pavlin-policar:
As per https://github.com/pavlin-policar/fastTSNE/issues/7#issuecomment-421938245, it is very important for optimization that the initialization has small variance e.g. 0.0001. Therefore in order to make PCA initialization usable, we should standardize the PCA projection.

PR #11, opened 2018-09-20
Ensure PCA initialization has small variance

pavlin-policar:
- As per #10, it is vital for tSNE optimization that the initialization has small variance. Without this scaling, PCA initialization can potentially lead to really bad results.

- Remove `pca` option when embedding new points into an existing embedding. This makes no sense because likely the points in the original space would have moved significantly and adding new points would be meaningless.

### DIFF
--- a/fastTSNE/nearest_neighbors.py
+++ b/fastTSNE/nearest_neighbors.py
@@ -44,8 +44,7 @@ class KNNIndex:
 
 class KDTree(KNNIndex):
     def build(self, data):
-        self.index = NearestNeighbors(
-            algorithm='kd_tree', metric=self.metric, n_jobs=self.n_jobs)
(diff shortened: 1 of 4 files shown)


dkobak:
Did you check if it makes a difference?

pavlin-policar:
I did - it makes a huge difference. Using non-scaled PCA initialization the results were very bizzare-looking, but scaling it down makes it nicely behaved. I also compared it against random initialization and it does seem like PCA does better. Random initialization seems like it would need more iterations to reach the same cluster separation, but this may be specific to the 5 tests I ran. Would definitely need to run more tests to verify this.

I will be preparing a couple of notebooks showing usage and I'll probably make one for this specifically. I don't think this is well known. Thanks for pointing it out!

PR #17, opened 2018-10-21
Add usage notebooks

pavlin-policar:
Add the following notebooks:
- prepare zeisel_2018 data set into a pickle
- simple usage
- simple transform
- advanced usage
- animation


dkobak:
Nice stuff!

Do you have any references for your procedure of mapping new cells to the fixed t-SNE? I found one (obscure?) reference, but I am wondering if you know other ones.

pavlin-policar:
Thanks!

> Do you have any references for your procedure of mapping new cells to the fixed t-SNE?

No, not really. I looked for references while implementing this, but didn't really find anything useful. The way it's implemented was the most straightforward and simplest to implement.

> I found one (obscure?) reference

I'd be interested in what you found - the current method is convenient, but could be improved.

PR #48, opened 2019-01-24
[WIP] faiss-cpu nearest neighbors search.

acopar:
##### Issue

Alternative nearest neighbor search algorithm with faiss library:
+ faster compared to pynndescent
+ does not require numba
+ exact search, so it may have slightly better results

The following issues need to be resolved before considering inclusion in the project:
- currently, it does not support pip install, which is required for macOS. There is no official faiss package for pip.
- the second step (gradient descent) appears to require more iterations compared to approximate nearest neighbor search.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


pavlin-policar:
Closing due to inactivity. Feel free to reopen if you want to continue working on this.

Issue #93, opened 2019-08-31
Add support for callable metrics


jgraving:
##### Expected behaviour
pynndescent supports passing callable metrics compiled with numba, so I would expect openTSNE to be able to support this
##### Actual behaviour
When I pass a numba-compiled callable metric, openTSNE throws a ValueError
```python
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-44-b188b4c74e96> in <module>()
----> 1 embedding_train = tsne.fit(X_train)

/home/jake/.local/lib/python3.6/site-packages/openTSNE/tsne.py in fit(self, X)
   1040 
   1041         """
-> 1042         embedding = self.prepare_initial(X)
   1043 
   1044         try:

/home/jake/.local/lib/python3.6/site-packages/openTSNE/tsne.py in prepare_initial(self, X)
   1118             metric_params=self.metric_params,
   1119             n_jobs=self.n_jobs,
-> 1120             random_state=self.random_state,
   1121         )
   1122 

/home/jake/.local/lib/python3.6/site-packages/openTSNE/affinity.py in __init__(self, data, perplexity, method, metric, metric_params, symmetrize, n_jobs, random_state)
    124         k_neighbors = min(self.n_samples - 1, int(3 * self.perplexity))
    125         self.knn_index, self.__neighbors, self.__distances = build_knn_index(
--> 126             data, method, k_neighbors, metric, metric_params, n_jobs, random_state
    127         )
    128 

/home/jake/.local/lib/python3.6/site-packages/openTSNE/affinity.py in build_knn_index(data, method, k, metric, metric_params, n_jobs, random_state)
    274             metric_params=metric_params,
    275             n_jobs=n_jobs,
--> 276             random_state=random_state,
    277         )
    278 

/home/jake/.local/lib/python3.6/site-packages/openTSNE/nearest_neighbors.py in __init__(self, metric, metric_params, n_jobs, random_state)
     11     def __init__(self, metric, metric_params=None, n_jobs=1, random_state=None):
     12         self.index = None
---> 13         self.metric = self.check_metric(metric)
     14         self.metric_params = metric_params
     15         self.n_jobs = n_jobs

/home/jake/.local/lib/python3.6/site-packages/openTSNE/nearest_neighbors.py in check_metric(self, *args, **kwargs)
    184             )
    185 
--> 186         return super().check_metric(*args, **kwargs)
    187 
    188     def build(self, data, k):

[...]

PR #94, opened 2019-09-02
Add support for callable metrics

jgraving:
##### Issue

See #93  

##### Description of changes
Added the option for callable metrics to the `openTSNE.nearest_neighbors` module. `NNDescent` checks if metric is compiled with `numba`, gives an informative warning, and then attempts to automatically compile it. If the function is already compiled or `NNDescent` is not being used then the callable is passed without any further checks. Docstrings have also been updated to match other docstrings with multiple types. If you'd like unit tests for this, then please provide guidance on how best to do this. 

##### Includes
- [X] Code changes
- [ ] Tests
- [X] Documentation

### DIFF
--- a/openTSNE/nearest_neighbors.py
+++ b/openTSNE/nearest_neighbors.py
@@ -56,14 +56,14 @@ class KNNIndex:
     def check_metric(self, metric):
         """Check that the metric is supported by the KNNIndex instance."""
         if callable(metric):
-            pass
+            return metric
         elif metric not in self.VALID_METRICS:
             raise ValueError(
                 f"`{self.__class__.__name__}` does not support the `{metric}` "
                 f"metric. Please choose one of the supported metrics: "
                 f"{', '.join(self.VALID_METRICS)}."
             )
-        return metric
+            return metric
 
 
 class BallTree(KNNIndex):
@@ -175,7 +175,7 @@ class NNDescent(KNNIndex):
         "yule",
     ]
 
-    def check_metric(self, metric, *args, **kwargs):
+    def check_metric(self, *args, **kwargs):
         import pynndescent
 
         if not np.array_equal(pynndescent.distances.named_distances, self.VALID_METRICS):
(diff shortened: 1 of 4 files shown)


pavlin-policar:
Thanks for the PR. I am missing tests, so we verify that regular python functions properly get compiled to numba code and a test with already numba compiled code.

Also, as mentioned in #93, it would be great if we could keep things consistent, and that you would add callable support to the BallTree as well.

jgraving:
I'll try adding tests, but I'm not very familiar and may need some guidance. `BallTree` doesn't overwrite the `openTSNE.nearest_neighbors.KNNIndex.check_metric` method, so support for callable metrics is already there.

jgraving:
Tests are added and passing. Let me know if you want me to add something more.

openTSNE/nearest_neighbors.py
```diff
@@ -183,7 +186,25 @@ def check_metric(self, *args, **kwargs):
                 "developers of this change."
             )
 
-        return super().check_metric(*args, **kwargs)
+        if callable(metric):
+            from numba.targets.registry import CPUDispatcher
+
+            if not isinstance(metric, CPUDispatcher):
+
```
pavlin-policar: no need for empty line

tests/test_nearest_neighbors.py
```diff
@@ -121,3 +179,21 @@ def test_random_state_being_passed_through(self, nndescent):
 
         nndescent.assert_called_once()
         check_mock_called_with_kwargs(nndescent, {"random_state": random_state})
+
+    def test_uncompiled_callable_is_compiled(self):
+
```
pavlin-policar: no need for empty line

tests/test_nearest_neighbors.py
```diff
@@ -65,6 +68,63 @@ def test_query_same_result_with_fixed_random_state(self):
...
 
+    def test_uncompiled_callable_metric_same_result(self):
+        k = 15
+
+        knn_index = self.knn_index("manhattan", random_state=1)
+        knn_index.build(self.x1, k=k)
+        true_indices_, true_distances_ = knn_index.query(self.x2, k=k)
+
+        def manhattan(x, y):
+            r"""Manhattan, taxicab, or l1 distance.
```
pavlin-policar: no need for docstring; it's not public and the code is self-explanatory.

tests/test_nearest_neighbors.py
```diff
@@ -65,6 +68,63 @@ def test_query_same_result_with_fixed_random_state(self):
...
+    def test_numba_compiled_callable_metric_same_result(self):
+        k = 15
+
+        knn_index = self.knn_index("manhattan", random_state=1)
+        knn_index.build(self.x1, k=k)
+        true_indices_, true_distances_ = knn_index.query(self.x2, k=k)
+
+        @njit(fastmath=True)
+        def manhattan(x, y):
+            r"""Manhattan, taxicab, or l1 distance.
```
pavlin-policar: no need for docstring; it's not public and the code is self-explanatory.

tests/test_nearest_neighbors.py
```diff
@@ -121,3 +179,21 @@ def test_random_state_being_passed_through(self, nndescent):
 
         nndescent.assert_called_once()
         check_mock_called_with_kwargs(nndescent, {"random_state": random_state})
+
+    def test_uncompiled_callable_is_compiled(self):
+
+        knn_index = nearest_neighbors.NNDescent("manhattan")
+
+        def manhattan(x, y):
+            r"""Manhattan, taxicab, or l1 distance.
```
pavlin-policar: no need for docstring; it's not public and the code is self-explanatory.

jgraving:
Should be good to go! Style errors are taken care of. I need to find a better linter. Have you seen black? https://black.readthedocs.io/en/stable/index.html I've been thinking about using it for my projects.

pavlin-policar:
Awesome, looks great!

> Have you seen black?

Yes, I'm a fan of black. Most of the code in `openTSNE` is compliant with its style guide, with the exception of cython files and the main `tsne.py`. The only reason I haven't implemented it fully is that it would introduce a whole lot more line breaks, which would IMO be less readable in this case. Otherwise, I try to use it wherever I can in my other projects.

pavlin-policar:
I forgot to squash your commits before merging, so I had to go through some git-gymnastics to get everything down to one coherent commit. That's why there's another commit in the history here. I didn't want to bother you with this, but next time, it's better that you squash your commits into a single one with a descriptive commit message, or do an interactive rebase to organize your commits. I've taken care of it now, though. Thanks for your contribution :smile:

jgraving:
Ah, sorry about that. I'll keep that in mind for next time. No problem! Glad I could help 😄

Issue #101, opened 2020-01-05
Runtime and RAM usage compared to FIt-SNE


dkobak:
I understand that openTSNE is expected to be slower than FIt-SNE, but I'd like to understand how much slower it is in typical situations. As I reported earlier, when I run it on 70000x50 PCA-reduced MNIST data with default parameters and `n_jobs=-1`, I get ~60 seconds with FIt-SNE and ~120 seconds with openTSNE. Every 50 iterations take around 2s vs around 4s.

I did not check for this specific case, but I suspect that FFT takes only a small fraction of this time, and the computational bottleneck is formed by the attractive forces. Can one profile openTSNE and see how much time is taken by different steps, such as repulsive/attractive computations?

Apart from that, and possibly even more worryingly, I replicated the data 6x and added some noise, to get a 420000x50 data matrix. It takes FIt-SNE around 1Gb of RAM to allocate the space for the kNN matrix, so it works just fine on my laptop. However, openTSNE rapidly took >7Gb of RAM and crashed the kernel (I have 16 Gb but around half was taken by other processes). This happened in the first seconds, so I assume it happens during the kNN search. Does pynndescent eat up so much memory in this case?

pavlin-policar:
> I did not check for this specific case, but I suspect that FFT takes only a small fraction of this time, and the computational bottleneck is formed by the attractive forces. Can one profile openTSNE and see how much time is taken by different steps, such as repulsive/attractive computations?

Yes, it's possible to benchmark this, but it's a little involved. You can add the `# cython: profile=True` directives to the cython files, then compile the thing from source by running `python setup.py develop`. Then you can get reports with cProfile, which is the default one if you're using pycharm. It's been a while since I've done any benchmarking, but from what I remember, neither of the attractive or repulsive forces "dominated" the runtime and I spent quite a bit optimizing the computation of the attractive forces to use as little instructions as possible.

I fear there isn't all that much I can do to make it faster since the generated C code from cython is a lot, and there is a lot of overhead to switching between compiled cython code and python compared to pure C++.

[...]

PR #111, opened 2020-03-10
Added Annoy support

dkobak:
Added Annoy support as per #101. Annoy is used by default if it supports the given metric and if the input data is not scipy.sparse (otherwise Pynndescent is used).

This needs installed Annoy (I installed this https://anaconda.org/conda-forge/python-annoy), but I wasn't sure where to add this dependency.


dkobak:
Wow, _all_ checks have failed :-/

Not sure what is going on here, maybe you could take a look...

pavlin-policar:
It looks like the test are failing because annoy is not being installed. Makes sense, as you didn't add it to any dependency list. Strict dependencies should be added to `setup.py`, while optional dependencies should not go there. Instead, if we introduce this as an optional dependency, add the following block to `azure-pipelines.yml`:

```diff
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -58,6 +58,10 @@ jobs:
   - bash: mv openTSNE src
     displayName: 'Remove source files from path'

+  - script: |
+      python -m pip install annoy
+    displayName: 'Install optional dependencies'
+
   - script: pytest -v
     timeoutInMinutes: 15
     displayName: 'Run unit tests'
```

If we want to add this as a hard dependency to replace pynndescent and have pynndescent be the optional dependency (which I am considering for the next release), it will need to be added to the conda recipe [here](https://github.com/conda-forge/opentsne-feedstock/blob/master/recipe/meta.yaml).

openTSNE/affinity.py
```diff
@@ -255,9 +255,17 @@ def check_perplexity(self, perplexity):
 def build_knn_index(
     data, method, k, metric, metric_params=None, n_jobs=1, random_state=None
 ):
+    if not sp.issparse(data) and metric in ["angular", "euclidean", "manhattan", "hamming", "dot"]:
```
pavlin-policar: Instead of introducing the distance naming conventions that annoy uses, I would prefer to have them mapped to the distance metrics we have now e.g. add support for _l1_ and _l2_, _angular_ is equivalent to cosine, I believe, and I'm not sure what _dot_ is useful for (assuming it's the dot product).

openTSNE/affinity.py
```diff
@@ -255,9 +255,17 @@ def check_perplexity(self, perplexity):
...
+        approxmethod = nearest_neighbors.Annoy
+    else:
+        approxmethod = nearest_neighbors.NNDescent
+
     methods = {
         "exact": nearest_neighbors.BallTree,
-        "approx": nearest_neighbors.NNDescent,
+        "auto": approxmethod,
```
pavlin-policar: `approxmethod` should be changed to `approx_method`

openTSNE/tsne.py
```diff
@@ -984,7 +986,7 @@ def __init__(
         min_grad_norm=1e-8,
         max_grad_norm=None,
         n_jobs=1,
-        neighbors="approx",
+        neighbors="auto",
```
pavlin-policar: `"auto"` is nice, but then it might be a good idea to have it switch between exact search and approximate search based on the number of neighbors. Otherwise it's more of _auto_for_approx_search_.

dkobak: As I wrote in some other discussion, the speed increase using exact search is well below 1s, so I thought it's not worth the effort. But I could make it default to `BallTree` for N<1000 (?) if you prefer.

openTSNE/affinity.py
dkobak: Yes, I believe "angular" is an alias for "cosine" that Annoy uses. I can map "cosine"->"angular" within the Annoy class. Should l1 and l2 be mapped to euclidean and manhattan? The "dot" is a bit weird, but Annoy explicitly supports it, so why not?

pavlin-policar: > Should l1 and l2 be mapped to euclidean and manhattan?

Yes. "taxicab" should also be mapped to manhattan, since this is what pynndescent supports. I'd like to keep consistency as much as possible.

We can also use cosine distance to calculate the pearson correlation, since it's just the cosine on centered data.

It should also be pretty easy to add the `dot` to pynndescent, so that that could also be used for sparse data.

openTSNE/tsne.py
pavlin-policar: Oh, I see, I must have missed that. In that case, it's safe to ignore it.

openTSNE/affinity.py
dkobak: OK, I'll add these alias mappings to the PR, but the support for additional metrics (correlation in annoy, dot in pynndescent) can be added later/separately, I think.

dkobak: ok done

openTSNE/affinity.py
dkobak: ok done

dkobak:
Ran through black, added Annoy to `azure-pipelines` and addressed your comments.

openTSNE/nearest_neighbors.py
```diff
@@ -146,6 +146,92 @@ def query(self, query, k):
...
+        super().__init__(*args, **kwargs)
+        self.__data = None
+        annoy_aliases = {
+            "cosine": "angular",
+            "l1": "manhattan",
+            "l2": "euclidean",
+            "taxicab": "manhattan",
+        }
+        if self.metric in annoy_aliases:
+            self.metric = annoy_aliases[self.metric]
```
pavlin-policar: Please don't change `self.metric` here, but change it to the correct string at the last minute in `build`. The reason for this is that this won't work if the user manually changes the metric after initialization. Also, if they want to check which metric was used, they may be in for some confusion: 

```python
index = Annoy(metric="cosine")
print(index.metric)  # will be angular, not cosine
```

openTSNE/affinity.py
```diff
@@ -255,9 +255,26 @@ def check_perplexity(self, perplexity):
 def build_knn_index(
     data, method, k, metric, metric_params=None, n_jobs=1, random_state=None
 ):
+    if not sp.issparse(data) and metric in [
```
pavlin-policar: This does not check anywhere if annoy is actually installed. I'd start with annoy being an optional dependency. This is easy as well, just add some flag at the top of the file like

```python
try:
    import annoy
    HAS_ANNOY = True
except ImportError:
    HAS_ANNOY = False
```

dkobak: Are you sure about that? Why not make it a compulsory dependency? It's on pip/conda, it's very stable, so there shouldn't be any problems. If you make it optional then most end users will never see the difference because they won't have it installed.

dkobak:
Why does it say that "build failed"?

pavlin-policar:
The tests are failing: see https://github.com/pavlin-policar/openTSNE/pull/111#pullrequestreview-373719742

openTSNE/affinity.py
pavlin-policar: I thought about gradually introducing it, in case there are some issues, and until we perform proper benchmarks. But either way, I just want to get this merged into master, so I can play with this on my end. In that case, please add `annoy` to the `setup.py` dependencies. I'll add it to the conda recipe before release.

dkobak: OK. I added it to setup.py and removed from azure-pipelines. You can change it if you prefer.

dkobak: I think it's called `annoy` in pip but `python-annoy` in conda, so I wasn't completely sure what to write in the `setup.py` though. I wrote `annoy`.

openTSNE/nearest_neighbors.py
dkobak: ok done

dkobak:
I feel stupid, but I cannot figure out how to read the failed test messages. E.g. It says `Test (Unit tests linux-python36) Failing after 3m`. When I click on "Details", I see ` Build log #L167. Bash exited with code '1'.` and not much else. How can I know which unit test failed?

pavlin-policar:
It's generally a good idea to run the tests locally on your machine first to see if they pass. You can do this easily with `python -m unittest tests/*.py`. You can see the azure logs if you press the _View more details on Azure Pipelines_ link below the _Bash exited with code '1'._ Then you can see that the task _Run unit tests_ is failing on all the different platforms.

openTSNE/affinity.py
pavlin-policar: That's right. Pypi dependencies, which are installed through pip have to go into either the setup.py file or sometimes the requirements.txt file. Conda packages have to be included in the conda recipe, which has its own repository (https://github.com/conda-forge/opentsne-feedstock), so `python-annoy` will go in there.

pavlin-policar:
From what I can see from the error logs, all you need to do is change `neighbors="approx"` to `neighbors="pynndescent"` in the failing tests, and they should pass.

dkobak:
Thanks! Running locally was indeed much more helpful. Fixed that, will try to add some tests now.

dkobak:
OK done. I ran the tests without changing the Annoy class first and everything passed, but I wasn't sure why the `test_query_train_same_result_with_fixed_random_state` actually passed because I did not set the seed of the AnnoyIndex. I fixed that anyway.

Also, I moved the two tests of callable metrics to the `TestNNDescent` as you suggested, but I'm not sure why it was passing before that: I think BallTree cannot take a callable or numba-compiled metric.

dkobak:
> I think BallTree cannot take a callable or numba-compiled metric.

Oh actually it can! I read this now in the #93 discussion.

In this case, should we copy those two tests into the `TestBallTree` as well?

pavlin-policar:
Yes, if it supports them, we should definitely test for that.

openTSNE/affinity.py
```diff
@@ -255,9 +257,26 @@ def check_perplexity(self, perplexity):
...
+        "cosine",
+        "euclidean",
+        "manhattan",
+        "hamming",
+        "dot",
+        "l1",
+        "l2",
+        "taxicab",
+    ]:
+        approx_method = nearest_neighbors.Annoy
```
pavlin-policar: `preferred_approx_method` is more descriptive.

openTSNE/affinity.py
```diff
@@ -255,9 +257,26 @@ def check_perplexity(self, perplexity):
...
+    else:
+        approx_method = nearest_neighbors.NNDescent
+
     methods = {
         "exact": nearest_neighbors.BallTree,
-        "approx": nearest_neighbors.NNDescent,
+        "auto": approx_method,
```
pavlin-policar: I know we've already discussed this before elsewhere, and you had pointed out that Annoy is just as fast as exact search for smaller data sets. Even so, wouldn't it be better to prefer exact search wherever possible just from the correctness point of view? Then "approx" might switch between annoy and nndescent, and "auto" between all three. This doesn't need to be fixed here and we can put this in once a cutoff is determined.

setup.py
```diff
@@ -268,6 +268,7 @@ def readme():
         "scikit-learn>=0.20",
         "scipy",
         "pynndescent>=0.4",
+        "annoy",
```
pavlin-policar: Please set this to be `annoy>=` whatever the version you have installed.

openTSNE/affinity.py
dkobak: Sure. How about I put a 1000 threshold for now, and you can adapt it later if you want? By the way, the same reasoning would suggest to implement exact t-SNE and actually default to that for very small datasets. I think the lack of exact t-SNE implementation (that I guess could be nicely parallelized too) is a shortcoming of openTSNE.

setup.py
dkobak: scipy does not have `>=` in here, is it on purpose?

pavlin-policar: scipy is required by a lot of packages and it's all too easy to cause requirement conflicts. Also, scipy is really good at back-compatibility.

openTSNE/affinity.py
pavlin-policar: That works. Yes, I have been thinking of adding exact t-SNE, but I'm still figuring out how to incorporate it into the current framework. This won't go in this release though.

setup.py
dkobak: ok done

openTSNE/affinity.py
dkobak: OK done. Great to hear that it's on the to-do!

openTSNE/affinity.py
dkobak: done

dkobak:
> Yes, if it supports them, we should definitely test for that.
Added that.

dkobak:
Codacy complaining about things like trailing whitespaces is driving me nuts...

pavlin-policar:
> Codacy complaining about things like trailing whitespaces is driving me nuts...

Hehe, I guess, but it's there to ensure code quality.

Also, I just noticed you were working on your master branch. It's generally best practice to create a separate branch and work on that, and make a PR on that branch. Just for future reference :)

dkobak:
> Also, I just noticed you were working on your master branch. It's generally best practice to create a separate branch and work on that, and make a PR on that branch. Just for future reference :)

All right. It's been a learning experience! :)

PR #113, opened 2020-03-12
Updates

### LINKED ISSUE #104: Change default learning rate for large data sets
See https://github.com/KlugerLab/FIt-SNE/issues/88.

### LINKED ISSUE #108: Non-consistent nan value for KL divergence
##### Expected behaviour

When calling `TSNE.fit()` on the same numpy array I expect:

1. The KL divergence to not be a `nan`
2. The appearence of `nan`in the KL divergence to not depend on ~~the specific  array slice considered~~ the chosen value of perplexity.

##### Actual behaviour

Parameters Tested (what is not mentioned is kept at the default value):
```python
perplexity = [100, 200, 300, 500]
distance = ["euclidean", "cosine"]
n_iter=2000
learning_rate=[200, X.shape[0] // 12]
n_jobs=4
```

Data Source:
```python
X = np.random.randint(-1, 1, shape=(600000, 100))
X_test_1 = X[:300000, :]
X_test_2 = X[:6000, :]
```
`X` is an embedding extracted using `tensorflow.keras.LSTM`.
`X` does not contain any "weird" value (e.g. `nan` or `inf`)

~~`X_test_1` leads to `nan` KL Divergence since the first 50 iterations~~
~~`X_test_2` shows normal behaviour~~
~~Random and bigger slices of  `X` also show normal behaviour~~

Values of perplexity above 100 leads to `nan` KL Divergence since the first 50 iterations

##### Steps to reproduce the behavior

It is hard to outline the steps for reproducing this behaviour since it appears to be data dependent. 
I have different behaviours on different slices of the input array while a colleague of mine reported a silent exit on one data source (of shape ~(10000, 250)) while normal behaviour on another.

**UPDATE**
The problem apparently lies in the parameter `perplexity`: values above 100 will lead to `nan` KL Divergence, is this behaviour expected?

### LINKED ISSUE #112: Add `verbose` input parameter
What about having `verbose` input parameter such that `verbose=1` would print the progress using the `ErrorLogger()` callback but also showing what happens before (kNN index construction, affinity calculation, etc.) and giving the times taken by the each stage?

pavlin-policar:
##### Description of changes
Range of minor updates.

- Fix initialization std from 1e-2 to 1e-4
- Change the default number of iterations to 750. Fixes #104 
- Add `auto` option for learning rate
- Replace the `ErrorLogger` with a `verbose` flag. Fixes #112 
- Remove `min_grad_norm` parameter. I feel like a deprecation warning here is unnecessary since I have never in my life seen that stopping condition hit, or used anywhere ever. 
- Fix KL divergence sometimes being reported as NaN. Fixes #108 

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


openTSNE/tsne.py
```diff
@@ -70,6 +70,10 @@ def _handle_nice_params(optim_params: dict) -> None:
 
     optim_params["n_jobs"] = n_jobs
 
+    # Determine learning rate if requested
+    if optim_params.get("learning_rate", "auto") == "auto":
+        optim_params["learning_rate"] = max(200, embedding.shape[0] / 12)
```
dkobak: Maybe change 12 to the early exaggeration factor here?

pavlin-policar: I wanted to do that, but then I realized, the way the "advanced" API is structured, this would only work properly for the early exaggeration phase since the calls to optimize are stateless.  So the learning rate during the early exaggeration phase would be N/12 and during the normal phase just N/1 (which I guess isn't right). This could work properly only in `TSNE.fit`.  So instead I opted for this, which should be okay in most cases. I don't think many people go around changing the default early exaggeration strength.

dkobak: OK makes sense.

dkobak:
George found what looks like a problem with the learning rate = n/12 heuristic: https://github.com/KlugerLab/FIt-SNE/pull/93

PR #115, opened 2020-03-15
Add spectral initialization using diffusion maps

### LINKED ISSUE #110: Spectral initialization
I think it'd be really cool to add `initialization='spectral'` support. It should be very easy, because it just means taking the P matrix after it's computed and running `scipy.sparse.linalg.svds` on it.

pavlin-policar:
##### Description of changes
Fixes #110.

I ended up implementing diffusion maps only because computationally, computing the leading eigenvectors is much faster than the smallest eigenvectors, and of the various spectral methods, diffusion maps are the only ones that require this. I checked what UMAP does - it uses the symmetric normalized laplacian for initialization - but they manually set the number of lanczos iteration limit, which I don't understand. This seemed like the better option.

@dkobak Do you want to take a look at this? I implemented this using `scipy.sparse.linalg.svds` because it turns out to be faster than `scipy.sparse.linalg.eigsh` and it seemed to produce strange results when I increased the error tolerance, while `svds` results seemed reasonable.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/openTSNE/initialization.py
+++ b/openTSNE/initialization.py
@@ -1,4 +1,5 @@
 import numpy as np
+import scipy.sparse as sp
 from sklearn.decomposition import PCA
 from sklearn.utils import check_random_state
 
@@ -66,6 +67,63 @@ def pca(X, n_components=2, svd_solver="auto", random_state=None):
     embedding /= normalization
 
     return np.ascontiguousarray(embedding)
+
+
+def spectral(A, n_components=2, tol=1e-4, max_iter=None):
+    """Initialize an embedding using the spectral embedding of the KNN graph.
+
+    Specifically, we initialize data points by computing the diffusion map on
+    the random walk transition matrix of the weighted graph given by the affiniy
+    matrix.
+
+    Parameters
+    ----------
+    A: Union[sp.csr_matrix, sp.csc_matrix, ...]
+        The graph adjacency matrix.
+
+    n_components: int
+        The dimension of the embedding space.
+
+    tol: float
+        See scipy.sparse.linalg.eigsh documentation.
+
+    max_iter: float
+        See scipy.sparse.linalg.eigsh documentation.
+
+    Returns
+    -------
+    initialization: np.ndarray
+
+    """
+    if A.ndim != 2:
+        raise ValueError("The graph adjacency matrix must be a 2-dimensional matrix.")
+    if A.shape[0] != A.shape[1]:
+        raise ValueError("The graph adjacency matrix must be a square matrix.")
+
+    D = sp.diags(np.ravel(np.sum(A, axis=1)))
+
+    # Find leading eigenvectors
+    k = n_components + 1
+    v0 = np.ones(A.shape[0]) / np.sqrt(A.shape[0])
+    eigvals, eigvecs = sp.linalg.eigsh(
+        A, M=D, k=k, tol=tol, maxiter=max_iter, which="LM", v0=v0
+    )
+    # Sort the eigenvalues in decreasing order
+    order = np.argsort(eigvals)[::-1]
+    eigvecs = eigvecs[:, order]
+
+    # In diffusion maps, we multiply the eigenvectors by their eigenvalues
+    eigvecs *= eigvals
+
+    # Drop the leading eigenvector
+    embedding = eigvecs[:, 1:]
+
+    # Ensure low variance
+    normalization = np.std(embedding[:, 0])
+    normalization /= 0.0001
+    embedding /= normalization
+
+    return embedding
 
 
 def weighted_mean(X, embedding, neighbors, distances):
--- a/openTSNE/tsne.py
+++ b/openTSNE/tsne.py
@@ -936,10 +936,10 @@ class TSNE(BaseEstimator):
 
     initialization: Union[np.ndarray, str]
         The initial point positions to be used in the embedding space. Can be a
-        precomputed numpy array, ``pca`` or ``random``. Please note that when
-        passing in a precomputed positions, it is highly recommended that the
-        point positions have small variance (var(Y) < 0.0001), otherwise you may
-        get poor embeddings.
+        precomputed numpy array, ``pca``, ``spectral`` or ``random``. Please
+        note that when passing in a precomputed positions, it is highly
+        recommended that the point positions have small variance
+        (std(Y) < 0.0001), otherwise you may get poor embeddings.
 
     metric: Union[str, Callable]
         The metric to be used to compute affinities between points in the
@@ -1122,6 +1122,18 @@ class TSNE(BaseEstimator):
             optimization.
 
         """
+
+        affinities = PerplexityBasedNN(
+            X,
+            self.perplexity,
+            method=self.neighbors_method,
+            metric=self.metric,
+            metric_params=self.metric_params,
+            n_jobs=self.n_jobs,
+            random_state=self.random_state,
+            verbose=self.verbose,
+        )
+
         # If initial positions are given in an array, use a copy of that
         if isinstance(self.initialization, np.ndarray):
             init_checks.num_samples(self.initialization.shape[0], X.shape[0])
@@ -1144,21 +1156,14 @@ class TSNE(BaseEstimator):
             embedding = initialization_scheme.random(
                 X, self.n_components, random_state=self.random_state
             )
+        elif self.initialization == "spectral":
+            embedding = initialization_scheme.spectral(
+                affinities.P, self.n_components, random_state=self.random_state
+            )
         else:
             raise ValueError(
                 f"Unrecognized initialization scheme `{self.initialization}`."
             )
-
-        affinities = PerplexityBasedNN(
-            X,
-            self.perplexity,
-            method=self.neighbors_method,
-            metric=self.metric,
-            metric_params=self.metric_params,
-            n_jobs=self.n_jobs,
-            random_state=self.random_state,
-            verbose=self.verbose,
-        )
 
         gradient_descent_params = {
             # Degrees of freedom of the Student's t-distribution. The


dkobak:
Yes, let me take a look.

> I ended up implementing diffusion maps only because computationally, computing the leading eigenvectors is much faster than the smallest eigenvectors, and of the various spectral methods, diffusion maps are the only ones that require this.

I'm not sure what you mean here; my understanding is that "diffusion maps" and "Laplacian eigenmaps" is exactly the same algorithm (mathematically equivalent), with the only difference being the scaling of eigenvectors by the respective eigenvalues [raised to some power in diffusion maps].

pavlin-policar:
Reading through @jlmelville's "Spectral methods", _Random Walk Normalized Laplacians_ are equivalent to _Diffusion Maps_. I may be using terminology wrong, but when I think of Laplacian eigenmaps, I think of the _Symmetrized Normalized Laplacian_ or just the _Unnormalized Graph Laplacian_. Now, unless I've got something really wrong, these are different from the Diffusion Maps since we're finding eigenvectors of different matrices.

dkobak:
> I implemented this using scipy.sparse.linalg.svds because it turns out to be faster than scipy.sparse.linalg.eigsh and it seemed to produce strange results when I increased the error tolerance, while svds results seemed reasonable.

What data did you test it on? When I tried your code snippet with line and circle, I got much better results using `eigsh` for generalized eigenvalue problem. `Svds` did not give me the correct largest eigenvector for the line. Maybe we can discuss it in #110, where the code snippet is posted.

Also, I remember that decreasing the tolerance to 1e-4 made eigsh not converge to the correct solution either.

I wonder if we can somehow use the knowledge that the largest eigenvectors is [1,1,...,1] with eigvenvalue=1 by construction... I noticed that you are providing `v0` to the `svds` call. I am wondering what happens if the first column is [1,1,1,1...]/sqrt(n) and the other columns are random but orthogonal to it. I didn't try it, it's just a wild idea.

dkobak:
> Reading through @jlmelville's "Spectral methods", Random Walk Normalized Laplacians are equivalent to Diffusion Maps. I may be using terminology wrong, but when I think of Laplacian eigenmaps, I think of the Symmetrized Normalized Laplacian or just the Unnormalized Graph Laplacian. Now, unless I've got something really wrong, these are different from the Diffusion Maps since we're finding eigenvectors of different matrices.

Symmetrized Normalized Laplacian is different. But Laplacian eigenmaps with Unnormalized Graph Laplacian (generalized eigenvector problem) = eigenvectors of Random Walk Normalized Laplacian = Diffusion Maps, as per https://jlmelville.github.io/smallvis/spectral.html#laplacian_eigenmaps section.

pavlin-policar:
I'd rather move the discussion here, so it's easy to look at the code changes.

I tested this on the macosko_2015 retina dataset, since I'd consider that a standard use-case. A reasonable number of points, it runs fast, basically, a very realistic example of where openTSNE might be used. The line and circle are fine as well, but they are very contrived examples, so I would place a lot less weight on those. 

Additionally, in real-world examples, the exact convergence really doesn't matter all that much, we just need an approximately good solution. It's going to be changed by t-SNE anyways. So I'm perfectly happy allowing larger error tolerances if it means it will run faster. And it seems to me that convergence for `eigsh` with `tol=1e-4` is much worse than for `svds`, e.g. on the full macosko dataset. Also, lobpcg and arpack give the same result.


```python
affinity = openTSNE.affinity.PerplexityBasedNN(x)
W = affinity.P
D = sp.diags(np.ravel(np.sum(W, axis=1)))
P = D.power(-1) @ W

U, S, VT = sp.linalg.svds(P, k=3, solver="arpack", tol=1e-4)
U1 = U[:, np.argsort(S)[::-1]]
S, U = sp.linalg.eigsh(P, 3, tol=1e-4)
U2 = U[:, np.argsort(S)[::-1]]

fig, ax = plt.subplots(ncols=2, figsize=(9, 4))
ax[0].scatter(U1[:, 1], U1[:, 2], c=y, s=1, alpha=0.05)
ax[0].set_title("arpack")
ax[1].scatter(U2[:, 1], U2[:, 2], c=y, s=1, alpha=0.05)
ax[1].set_title("eigsh")
```

image

Also, I ran some benchmarks on the three methods on different samples of the macosko data set and these are the results, 10 repetitions for each sample:

```python
results = []

sizes = (100, 1000, 5000, 10000, 20000, 30000, 40000)
for s in sizes:
    print("size", s)
    indices = np.random.choice(range(x.shape[0]), size=s)
    sample = x[indices]

    affinity = openTSNE.affinity.PerplexityBasedNN(x)
    W = affinity.P
    D = sp.diags(np.ravel(np.sum(W, axis=1)))
    P = D.power(-1) @ W

    for i in range(10):
        start = time()
        U, S, VT = sp.linalg.svds(P, k=3, solver="arpack", tol=1e-4)
        results.append(("arpack", s, time() - start))

    for i in range(10):
        start = time()
        U, S, VT = sp.linalg.svds(P, k=3, solver="lobpcg", tol=1e-4)
        results.append(("lobpcg", s, time() - start))

    for i in range(10):
        start = time()
        S, U = sp.linalg.eigsh(P, 3, tol=1e-4)
        results.append(("eigsh", s, time() - start))

df = pd.DataFrame(results, columns=["method", "size", "time"])
ax = sns.lineplot(data=df, x="size", y="time", hue="method")
ax.set_ylim(0, ax.get_ylim()[1])
```

image

`eigsh` is 2x slower than `arpack`. Granted, it is only 4 seconds compared to 2 seconds, and that seems very small, but combined with the worse convergence for higher tolerance rates, and I really don't see any reason to not use `svds`.

dkobak:
Wait - that's not how I think you should be running `eigsh`. It should be 
```
S,U = sp.linalg.eigsh(W, M=D, k=3)
```
because you want a generalized eigenvalue problem with W and D. The P is not symmetric and so `eisgh` should not give meaningful results there.

dkobak:
Apart from that, I am really curious about the effect of starting the iterations with this `v0`:
```
n = 10000
v0 = np.random.randn(n,3)
v0[:,0] = np.ones(n) / np.sqrt(n) # the first eigenvector is [1,1,...1]
v0,_ = np.linalg.qr(v0)           # orthogonalizing the columns
v0[:,0] = np.ones(n) / np.sqrt(n) # in case the sign was flipped
```

pavlin-policar:
So the affinity matrix P is symmetric because we explicitly symmetrize W for t-SNE, and the degree matrix is diagonal and, hence, also symmetric. 

I don't know very much about generalized eigenvalue problems and such, but running `eigsh` your way is even slower than the way I ran it, and it gives visually similar results, which can probably be attributed to the differences in random `v0` initialization since we leave it random.

image

Curiously enough, arpack is slower than eigsh for both the line and circle. Line: arpack 7.4sec, eigsh 1.6s and circle: arpack 8.8s, eigsh 1.3s. Both line and circle were generated with 10k points. It's kind of puzzling to me, but maybe because they are such contrived examples, svds is having a more difficult time converging?

As for the `v0` initialization: The reason I provide my own initialization is that `svds` doesn't allow us to specify a seed, and it creates v0 randomly. If I increase error tolerance, then v0 might affect the result, and the only way to have reproducible results is to initialize it ourselves. I'm not entirely sure what using the initialization you want would achieve? Do you think it would converge faster if we provide an orthonormal matrix? From my understanding, most power iteration methods already do this after every iterations, but I have no idea how arpack, lobpcg, or eigsh work, so I don't know. I'm not really sure what you want to achieve with this v0 initialization.

dkobak:
> So the affinity matrix P is symmetric because we explicitly symmetrize W for t-SNE, and the degree matrix is diagonal and, hence, also symmetric.

A product of diagonal and symmetric matrix is not necessarily symmetric. Try multiplying [1, 0; 0, 3] by [2, 1; 1, 3].

> I don't know very much about generalized eigenvalue problems and such, but running eigsh your way is even slower than the way I ran it, and it gives visually similar results

OK, it is possible that somehow they give similar result on these examples, but only "my" way is valid :-)

> I'm not entirely sure what using the initialization you want would achieve? Do you think it would converge faster if we provide an orthonormal matrix?

No, this isn't the idea. The idea is that we KNOW the first eigenvector (the one that we are discarding). It is [1,1,1.... 1] (with eigenvalue 1). So svds/eigsh do not need to spend time looking for this eigenvector, because we already know it in advance. I thought that initializing v0 with the correct leading eigenvector would make the results more accurate (especially with high tolerance) and possibly also faster.

dkobak:
If you say that `eigsh` done my way on the Macosko data looks visually similar, then I am confused about why `svds` and `eigsh` give so hugely different results (visually at least they look completely different). Could you run both with `tol=0` and show the resulting scatter plots?

pavlin-policar:
> A product of diagonal and symmetric matrix is not necessarily symmetric. Try multiplying [1, 0; 0, 3] by [2, 1; 1, 3].

Oh, yeah, of course, you're right. That was silly of me.

> If you say that eigsh done my way on the Macosko data looks visually similar, then I am confused about why svds and eigsh give so hugely different results (visually at least they look completely different). Could you run both with tol=0 and show the resulting scatter plots?

My bad here, I forgot to do `D.power(-1)`. Now the results are very similar:
image

Interestingly, running eigsh your way without inverting the matrix gives very similar results to my improper call to eigsh... So that's weird.

> No, this isn't the idea. The idea is that we KNOW the first eigenvector (the one that we are discarding). It is [1,1,1.... 1] (with eigenvalue 1). So svds/eigsh do not need to spend time looking for this eigenvector, because we already know it in advance. I thought that initializing v0 with the correct leading eigenvector would make the results more accurate (especially with high tolerance) and possibly also faster.

Oh, that's a great idea. Is that always the case though?

Another thing I've forgotten is to multiply the eigenvectors by the eigenvalues...

pavlin-policar:
Rerunning the benchmarks, calling `eigsh` with the proper matrix, this is now faster than `svds`.

image

This is the way to go then.

dkobak:
> My bad here, I forgot to do D.power(-1). Now the results are very similar:

Just to double check: the call should be `sp.linalg.eigsh(W, M=D)` where `D` is the row sums of W, without any inversion. Is that what you did for the third panel of the figure you posted?

And does this figure show the results with `tol=0` or `tol=1e-4`?

pavlin-policar:
So, I am inverting the degree matrix now, and I wasn't before. Inverting it is correct because we're trying to find the eigenvectors of P = D^{-1} W. Here's the entire code:

```python
W = affinity.P
D = sp.diags(np.ravel(np.sum(W, axis=1))) # no inversion here
P = D.power(-1) @ W

U, S, VT = sp.linalg.svds(P, k=3, solver="arpack", tol=1e-4)
U1 = U[:, np.argsort(S)[::-1]]
S, U = sp.linalg.eigsh(P, 3, tol=1e-4)
U2 = U[:, np.argsort(S)[::-1]]
# Forgot to invert D below in previous images
S, U = sp.linalg.eigsh(W, M=D.power(-1), k=3, tol=1e-4)
U3 = U[:, np.argsort(S)[::-1]]

fig, ax = plt.subplots(ncols=3, figsize=(13, 4))
ax[0].scatter(U1[:, 1], U1[:, 2], c=y, s=1, alpha=0.05)
ax[0].set_title("arpack")
ax[1].scatter(U2[:, 1], U2[:, 2], c=y, s=1, alpha=0.05)
ax[1].set_title("eigsh")
ax[2].scatter(U3[:, 1], U3[:, 2], c=y, s=1, alpha=0.05)
ax[2].set_title("eigsh (with M=D)")
```

pavlin-policar:
It kind of seems like initializing v0 to 1/sqrt(N) might help, but the effect is very minor:
image
However, I can only provide the initial vector, not matrix, as your code does, so the entire code ends up being:

```python
v0 = np.ones(n) / np.sqrt(n)
S, U = sp.linalg.eigsh(W, M=D.power(-1), k=3, tol=1e-4, v0=v0)
```

dkobak:
> Inverting it is correct because we're trying to find the eigenvectors of P = D^{-1} W

No, it isn't correct. We are trying to solve generalized eigenvalue problem with W and D. Without inversion.

> tol=1e-4

I thought that looking at the results with tol=0 should give us an idea about how well this is converged.

dkobak:
I think what we should see in the outcome is a good separation of the cell types. This is only visible in the `eigsh` plot without inversion.

pavlin-policar:
> No, it isn't correct. We are trying to solve generalized eigenvalue problem with W and D. Without inversion.

Ok, that's because we left multiply the expression

> D^{-1}A @ x = w* x

with D to get

> A @ x = w * D @ x

which is what `eigsh` requires. Right?

Also, if the matrix P is not symmetric, then the SVD is not equivalent to the eigendecomposition since that only holds for real, symmetric matrices. Does that make sense?

So if all that holds, then using `svds` is just plain wrong.

dkobak:
> Ok, that's because we left multiply the expression [...] which is what eigsh requires. Right?

Yes.

> Also, if the matrix P is not symmetric, then the SVD is not equivalent to the eigendecomposition since that only holds for real, symmetric matrices. Does that make sense? So if all that holds, then using svds is just plain wrong.

Hmm yes, I did not think of that, but I think you are right.

So it's either `eigs(P)` or `eigs(W, M=D)` or `eigsh(W, M=D)`. The eigsh assumes that W is symmetric so I think should be either faster or more accurate than eigs(W, M=D)`.

Okay, now that this is clarified, let's check how `eigsh(W,M=D)` compares to `eigs(P)` with `tol=1e-4` and with `tol=0` and possibly with/without `v0=v0`...

dkobak:
Hey, I re-read this thread and noticed that you already showed the results with `tol=0`. I think we are on the same page now. I like the version in the current PR that uses `eigsh(W,M=D,tol=1e-4,v0=v0)`.

One small thing that I didn't reply to:

> Oh, that's a great idea. Is that always the case though?

Yes, [1,1...1]  will always be the leading eigenvector because P has constant row sums.

I am a little bit concerned about `tol=1e-4`. It works fine for Macosko but it will cause problems like described here: https://github.com/lmcinnes/umap/issues/360. For the line and the circle `tol=1e-4` fails to recover the correct eigenvectors, but `tol=0` works well. Unfortunately, it does take a long time: I get 35 sec vs. 1 sec for a line with n=10000.

Screenshot from 2020-03-16 21-07-05

pavlin-policar:
Here we are. `eigs` vs `eigsh` with `tol=0` and `tol=1e-4`. I realized that I wasn't actually running the benchmark on samples in the above plots, that's why the curves were flat above.

<details><summary>Code</summary>

```python
results = []
n_reps = 5

sizes = (1000, 5000, 10000, 25000, 50000, 75000)
for s in sizes:
    print("size", s)
    indices = np.random.choice(range(x.shape[0]), size=s)
    sample = x[indices]

    affinity = openTSNE.affinity.PerplexityBasedNN(sample)
    W = affinity.P
    D = sp.diags(np.ravel(np.sum(W, axis=1)))
    P = D.power(-1) @ W

    for i in range(n_reps):
        v0 = np.random.randn(W.shape[0])
        start = time()
        S, U = sp.linalg.eigsh(W, M=D, k=3, tol=0, v0=v0)
        results.append(("eigsh (tol=0)", s, time() - start))

    for i in range(n_reps):
        v0 = np.ones(W.shape[0]) / np.sqrt(W.shape[0])
        start = time()
        S, U = sp.linalg.eigsh(W, M=D, k=3, tol=1e-4, v0=v0)
        results.append(("eigsh (tol=1e-4)", s, time() - start))

    for i in range(n_reps):
        v0 = np.random.randn(W.shape[0])
        start = time()
        S, U = sp.linalg.eigs(P, k=3, tol=0, v0=v0)
        results.append(("eigs (tol=0)", s, time() - start))

    for i in range(n_reps):
        v0 = np.ones(W.shape[0]) / np.sqrt(W.shape[0])
        start = time()
        S, U = sp.linalg.eigs(P, k=3, tol=1e-4, v0=v0)
        results.append(("eigs (tol=1e-4)", s, time() - start))

df = pd.DataFrame(results, columns=["method", "size", "time"])
ax = sns.lineplot(data=df, x="size", y="time", hue="method")
ax.set_ylim(0, ax.get_ylim()[1])
ax.legend(bbox_to_anchor=(1.1, 1))
```
</details>

image

Increasing tolerance is clearly very beneficial, however, it seems that `eigs` is actually faster than `eigsh`, which is strange. I would have expected `eigsh` to be able to explit the fact that it's a symmetric matrix in some way or another. I ran some more benchmarks with just `tol=1e-4` with larger data sets by oversampling the macosko data set. I am assuming here that having duplicate entries will not affect the algorithms too much.

<details><summary>Code</summary>

```python
results = []
n_reps = 5

sizes = (1000, 50000, 100_000, 200_000, 300_000, 400_000, 500_000)
for s in sizes:
    print("size", s)
    indices = np.random.choice(range(x.shape[0]), size=s)
    sample = x[indices]

    affinity = openTSNE.affinity.PerplexityBasedNN(sample, n_jobs=-1)
    W = affinity.P
    D = sp.diags(np.ravel(np.sum(W, axis=1)))
    P = D.power(-1) @ W

    for i in range(n_reps):
        v0 = np.ones(W.shape[0]) / np.sqrt(W.shape[0])
        start = time()
        S, U = sp.linalg.eigsh(W, M=D, k=3, tol=1e-4, v0=v0)
        results.append(("eigsh (tol=1e-4)", s, time() - start))

    for i in range(n_reps):
        v0 = np.ones(W.shape[0]) / np.sqrt(W.shape[0])
        start = time()
        S, U = sp.linalg.eigs(P, k=3, tol=1e-4, v0=v0)
        results.append(("eigs (tol=1e-4)", s, time() - start))

df = pd.DataFrame(results, columns=["method", "size", "time"])
ax = sns.lineplot(data=df, x="size", y="time", hue="method")
ax.set_ylim(0, ax.get_ylim()[1])
ax.legend(bbox_to_anchor=(1.1, 1))
```
</details>


image

That's weird, it seems that `eigs` is actually faster than `eigsh`, but they're equal at 300k points, so that's wholly confusing. It seems like `eigs` is the better option.

pavlin-policar:
Here we are. `eigs` vs `eigsh` with `tol=0` and `tol=1e-4`. I realized that I wasn't actually running the benchmark on samples in the above plots, that's why the curves were flat above.

<details><summary>Code</summary>

```python
results = []
n_reps = 5

sizes = (1000, 5000, 10000, 25000, 50000, 75000)
for s in sizes:
    print("size", s)
    indices = np.random.choice(range(x.shape[0]), size=s)
    sample = x[indices]

    affinity = openTSNE.affinity.PerplexityBasedNN(sample)
    W = affinity.P
    D = sp.diags(np.ravel(np.sum(W, axis=1)))
    P = D.power(-1) @ W

    for i in range(n_reps):
        v0 = np.random.randn(W.shape[0])
        start = time()
        S, U = sp.linalg.eigsh(W, M=D, k=3, tol=0, v0=v0)
        results.append(("eigsh (tol=0)", s, time() - start))

    for i in range(n_reps):
        v0 = np.ones(W.shape[0]) / np.sqrt(W.shape[0])
        start = time()
        S, U = sp.linalg.eigsh(W, M=D, k=3, tol=1e-4, v0=v0)
        results.append(("eigsh (tol=1e-4)", s, time() - start))

    for i in range(n_reps):
        v0 = np.random.randn(W.shape[0])
        start = time()
        S, U = sp.linalg.eigs(P, k=3, tol=0, v0=v0)
        results.append(("eigs (tol=0)", s, time() - start))

    for i in range(n_reps):
        v0 = np.ones(W.shape[0]) / np.sqrt(W.shape[0])
        start = time()
        S, U = sp.linalg.eigs(P, k=3, tol=1e-4, v0=v0)
        results.append(("eigs (tol=1e-4)", s, time() - start))

df = pd.DataFrame(results, columns=["method", "size", "time"])
ax = sns.lineplot(data=df, x="size", y="time", hue="method")
ax.set_ylim(0, ax.get_ylim()[1])
ax.legend(bbox_to_anchor=(1.1, 1))
```
</details>

image

It looks like both `eigs` and `eigsh` are running at the same speed with higher tolerance. I'll just keep the PR as is.

> I am a little bit concerned about tol=1e-4. It works fine for Macosko but it will cause problems like described here: lmcinnes/umap#360. For the line and the circle tol=1e-4 fails to recover the correct eigenvectors, but tol=0 works well. Unfortunately, it does take a long time: I get 35 sec vs. 1 sec for a line with n=10000.

I'm not particularly worried about this. If the default tolerance isn't good enough, it should be easy enough to call `initialization.spectral(..., tol=0)` and pass that as initialization. And this should only really be a problem with contrived examples.

If you're okay with the current state, I'll merge.

dkobak:
Sounds good. Go ahead & merge.

> If the default tolerance isn't good enough, it should be easy enough to call `initialization.spectral(..., tol=0)` and pass that as initialization. And this should only really be a problem with contrived examples.

OK. My only suggestion would be to add this to the docs with a code snipped showing how exactly to do it.

PR #118, opened 2020-03-17
Fixup verbose messages

pavlin-policar:
##### Description of changes
Fixup verbose messages.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


dkobak:
Cool! Out of curiosity -- is there anything else you are planning for the next release?

pavlin-policar:
The only thing remaining is to run a series of benchmarks to determine a good switching point between exact search and approx search using the `neighbors="auto"` setting. You had set it to 1k in the PR, but that seems largely arbitrary.

PR #122, opened 2020-04-19
Implements variable degrees of freedom in kernel

### LINKED ISSUE #51: Allow to vary the t-distribution degree of freedom
Hi Pavlin. Have you already seen https://arxiv.org/abs/1902.05804 by any chance? I thought you might be interested to add a `df` parameter to openTSNE. It was really easy to implement it in FIt-SNE so I guess it should be pretty easy here as well. I would even consider doing it myself, but alas, I am _so_ caught in revisions right now (including revising this paper) that I am not sure when I would have time for it...

Anyway, if you don't want to implement it (at least not until our paper is out in some reputable place haha) then it's fine. Do let me know if you have any comments on the preprint though.

pavlin-policar:
##### Description of changes
Fixes #51 and implements variable degrees of freedom in the Cauchy kernel.

@dkobak I finally took a few hours and implemented this. I also spent a bit of time trying to come up with an expression that would use only a single kernel. I was not successful. Did you guys try to do this when you were working on this? The performance impact is minimal, so it doesn't really matter, but I had to change the code up a bit to accommodate this.

I do like the plots with dof<1. I have grown to like the compactness characteristic of UMAP, so I tend to use exaggeration, and this lines up with that as well. Is there a reason to use prefer degrees of freedom over exaggeration? 

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation


dkobak:
Wow, Pavlin, this is very cool! Happy to see it implemented in openTSNE.

By the way -- does openTSNE support embedding into 1D? George has only recently implemented df!=1 support for 1D embeddings in FIt-SNE. This was picked up and used here: https://arxiv.org/abs/2002.05687.

> I also spent a bit of time trying to come up with an expression that would use only a single kernel. I was not successful. Did you guys try to do this when you were working on this? 

I'm actually not quite sure what you mean. Do you mean that you could not use the same kernel for df=1 and df!=1? Why? If I remember correctly, George did code it as separate functions in FIt-SNE but this was because he thought the code for df!=1 might work slower even if one passes in df=1. Not sure this was actually the case in the end.

> I do like the plots with dof<1. I have grown to like the compactness characteristic of UMAP, so I tend to use exaggeration, and this lines up with that as well. Is there a reason to use prefer degrees of freedom over exaggeration?

I think these two things have generally different effects. df<1 tends to yield more clusters ("finer cluster structure" as we call it in our paper), whereas exaggeration>1 tends to shrink clusters. Here I have a figure with MNIST, showing an effect of exaggeration=4 and df=0.5:

tsne-exagg-tails

The UMAP of MNIST looks just like exaggeration=4. Actually the default similarity kernel in UMAP does have a bit heavier tails than df=1, so I think maybe setting slightly lower df gets one a bit closer to the default UMAP, but the effect of exaggeration is stronger.

pavlin-policar:
> By the way -- does openTSNE support embedding into 1D?

Yes. And it's fully integrated into openTSNE so it works with `transform` and it should be compatible with everything else.

> Do you mean that you could not use the same kernel for df=1 and df!=1? Why? If I remember correctly, George did code it as separate functions in FIt-SNE but this was because he thought the code for df!=1 might work slower even if one passes in df=1. Not sure this was actually the case in the end.

No, we couldn't use the same kernels. So your paper doesn't go into detail about the FIt-SNE implementation, and that's fine -- deriving the things I needed was really straightforward. For their paper (https://www.nature.com/articles/s41592-018-0308-4), one possible setting of the phi terms is described, which needs two different kernels (K1 and K2 in eqn. (3)), but this is not what is actually implemented in FIt-SNE. @linqiaozhi came up with an alternative way so that he only needed to use K2.

So the same thing happens here. When I worked out the FIt-SNE approximation, I came up with two kernels, and I couldn't find a way to squish them into a single one. Not that it matters since the performance impact seems small. What might matter is that using a single kernel uses up less memory. I was wondering if you guys had tried to find a way to use a single kernel. Looking at the FIt-SNE implementation, it looks like we came to the same thing though.

The df != 1 is definitely slower because there's a whole lot of extra division and exponentiation operations. There's no way around that. Separating the two is definitely a good idea. I just incorporated it into the existing code using if statements. The CPU branch predictor should have no problems with that.

>I think these two things have generally different effects. df<1 tends to yield more clusters ("finer cluster structure" as we call it in our paper), whereas exaggeration>1 tends to shrink clusters.

Yeah, I have yet to try it on larger data sets, I'm really curious to see what I get there.

dkobak:
> but this is not what is actually implemented in FIt-SNE. @linqiaozhi came up with an alternative way so that he only needed to use K2.

Oh yes, true. Too bad it's not explicitly written down in the FIt-SNE paper.

> So the same thing happens here. When I worked out the FIt-SNE approximation, I came up with two kernels, and I couldn't find a way to squish them into a single one. Not that it matters since the performance impact seems small. What might matter is that using a single kernel uses up less memory. I was wondering if you guys had tried to find a way to use a single kernel. Looking at the FIt-SNE implementation, it looks like we came to the same thing though.

Yes, now I remember how George was explainining this to me. He had exactly the same problem and also decided to use two kernels. 

> The df != 1 is definitely slower because there's a whole lot of extra division and exponentiation operations. There's no way around that.

My feeling is that the difference might be too small to care about, but I did not do thorough testing.

linqiaozhi:
Glad to see variable dof implemented in openTSNE!

And yes, as Dmitry mentioned, we use two kernels for the implementation that allows variable dof. To be honest, I forget if I spent much time working it out for a single kernel or not...but ultimately I remember that we decided it was likely not worth it from a performance standpoint.

PR #133, opened 2020-05-14
max update norm parameter

### LINKED ISSUE #132: Limit max_step_norm
I'd like to open a discussion about the `max_step_norm` parameter that we introduced into the last release of FIt-SNE as a hack after George came up with an example where some points were shooting off very fast.

Since then, I ran into two other situations where the same thing happened in openTSNE. One example was running some experiments with very large exaggeration factor (30) and learning rate set to N/30. Some points were shooting off. Another example I had right now, using sparse data and cosine metric. I only had N=10000 so I ran Barnes-Hut and saw that several points ended up ~1000 distance away from the main embedding. Of course FFT just got stuck.

I inserted the following lines before `embedding += update` and it seems to have eliminated these issues:

```
            # clip the step sizes
            max_step_norm = 5
            update_norms = np.linalg.norm(update, axis=1, keepdims=True)
            norm_exceeded = update_norms.squeeze() > max_step_norm
            update[norm_exceeded] /= update_norms[norm_exceeded]
            update[norm_exceeded] *= max_step_norm
```

Also, I noticed UMAP does a very similar clipping of the step sizes (I assume for similar reasons).

You currently have `max_grad_norm` which is sensible, but the gradient later gets modified by the gains, momentum, and learning rate. On reflection, I think that limiting the `update` norms makes more sense than limiting the `gradient` norms.

dkobak:
Fixes #132.

I tested it on MNIST with default parameters and the default `max_step_norm=5` (same as in FIt-SNE) does not affect the embedding (I think it's never triggered). If one sets it to a much smaller number, it produces a funny effect:

tmp-opentsne-step001

Here with `max_step_norm=0.01` no point can be further than 7.5 from zero (with 750 iterations).

@jnboehm tested it on one case where we had some points shooting off (with strong exaggeration factor), and the default `max_step_norm=5` resolved the problem.

For `transform()` I set the default to `None` because there you have nonzero `max_grad_norm` by default so I thought it might be enough. You are also limiting the interpolation grid there, I think.

### DIFF
--- a/openTSNE/tsne.py
+++ b/openTSNE/tsne.py
@@ -607,6 +607,11 @@ class TSNEEmbedding(np.ndarray):
             the embedding, causing the interpolation method to compute a very
             large grid, and leads to worse results.
 
+        max_step_norm: float
+            Maximum update norm. If the norm exceeds this value, it will be
+            clipped. This prevents points from "shooting off" from
+            the embedding.
+
         random_state: Union[int, RandomState]
             The random state parameter follows the convention used in
             scikit-learn. If the value is an int, random_state is the seed used
@@ -690,6 +695,7 @@ class TSNEEmbedding(np.ndarray):
         initial_momentum=0.5,
         final_momentum=0.8,
         max_grad_norm=0.25,
+        max_step_norm=None,
     ):
         """Embed new points into the existing embedding.
 
@@ -762,6 +768,11 @@ class TSNEEmbedding(np.ndarray):
             the embedding, causing the interpolation method to compute a very
             large grid, and leads to worse results.
 
+        max_step_norm: float
+            Maximum update norm. If the norm exceeds this value, it will be
+            clipped. This prevents points from "shooting off" from
+            the embedding.
+
         Returns
         -------
         PartialTSNEEmbedding
@@ -798,6 +809,7 @@ class TSNEEmbedding(np.ndarray):
                 inplace=True,
                 propagate_exception=True,
                 max_grad_norm=max_grad_norm,
+                max_step_norm=max_step_norm,
             )
             embedding.optimize(
                 n_iter=n_iter,
@@ -807,6 +819,7 @@ class TSNEEmbedding(np.ndarray):
                 inplace=True,
                 propagate_exception=True,
                 max_grad_norm=max_grad_norm,
+                max_step_norm=max_step_norm,
             )
 
         except OptimizationInterrupt as ex:
@@ -1035,6 +1048,11 @@ class TSNE(BaseEstimator):
         the embedding, causing the interpolation method to compute a very
         large grid, and leads to worse results.
 
+    max_step_norm: float
+        Maximum update norm. If the norm exceeds this value, it will be
+        clipped. This prevents points from "shooting off" from
+        the embedding.
+
     n_jobs: int
         The number of threads to use while running t-SNE. This follows the
         scikit-learn convention, ``-1`` meaning all processors, ``-2`` meaning
@@ -1092,6 +1110,7 @@ class TSNE(BaseEstimator):
         initial_momentum=0.5,
         final_momentum=0.8,
         max_grad_norm=None,
+        max_step_norm=5,
         n_jobs=1,
         neighbors="auto",
         negative_gradient_method="fft",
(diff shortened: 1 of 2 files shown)


pavlin-policar:
Awesome, thanks!

That's a pretty neat looking effect and reminds me a bit of Poincare maps. Not sure how useful it might be in the context of t-SNE though.

> For transform() I set the default to None because there you have nonzero max_grad_norm by default so I thought it might be enough. You are also limiting the interpolation grid there, I think.

Yeah, that's good. Points off-shooting aren't a problem there anymore since now I just keep the grid static and don't allow points to move further away. This makes transform both super fast and visually more appealing IMO when the points don't match the training data. For example, this is the Cao data put onto the Tasic data. Not very informative, but looks really cool, I think.

image

pavlin-policar:
Could you just add this to the [`TestTSNEParameterFlow `](https://github.com/pavlin-policar/openTSNE/blob/master/tests/test_tsne.py#L81), so that it tests that the parameters are properly passed to the underlying functions? Add something like `"max_grad_norm": [None, 1, 5]` [here](https://github.com/pavlin-policar/openTSNE/blob/master/tests/test_tsne.py#L84-L95) and [here](https://github.com/pavlin-policar/openTSNE/blob/master/tests/test_tsne.py#L157-L163). Because the parameters have to be copied and delegated into so many objects, this checks that things are called with the proper params.

dkobak:
> For example, this is the Cao data put onto the Tasic data. 

Wow this looks really cool! I've never tried it with these datasets. Does the arrangement make sense? If you plot this Tasic embedding on its own with Tasic colors and the Cao overlay on its own with Cao colors (and maybe also cluster numbers), then one can check if it maps interneurons to interneurons etc. Would be cool to see if it can separate Sst and Pvalb populations, etc.

What metric did you use for `transform`?

pavlin-policar:
So looking at how I actually generated this example a while back, this only contains inhibitory neurons and some mesoderm cells from Cao (clusters 15 and 3). The inhibitory neurons kind of line up with the ones from tasic, so these are the orange ones.

image

image
image

I also just now mapped the Cao inhibitory interneurons as well, but the expression of Sst and Pvalb is really low in Cao. My guess is that since Cao has cells from a developing mouse (P11-P13) and tasic has cells from a full-grown mouse (P50-P90). The mapping isn't all that meaningful IMO, and I'd attribute that to developmental differences. If I take a data set with similar conditions to the Tasic data set, I can get very meaningful overlapping regions.

Also, I can merge as soon as you add the test.

dkobak:
Nice. Thanks for the additional plots. So where does the cluster 27 (inhibitory interneurons) end up mostly? To be honest, I don't understand what's the difference between 15 and 27 supposed to be.

pavlin-policar:
Yeah, so here is the reference.

image

My understanding of this stuff is very basic, but the Gads are the main thing needed for gabaergic neurons, which are inhibitory. I assume Sst and Pval are markers for inhibitory interneurons, so I guess there are other types of inhibitory neurons other than interneurons. So the bottom clusters here in Tasic are interneurons, and the top ones are something else, but still inhibitory. The top-other has the Lamp cluster and the bottom-other has the Vis clusters.

Here's where cluster 27 maps to
image
It definitely prefers the other inhibitory cluster that isn't interneurons, while some of it seems all over the place. Also, to answer your question from before, I did the transform with cosine similarity with 1000 genes.

dkobak:
> I assume Sst and Pval are markers for inhibitory interneurons, so I guess there are other types of inhibitory neurons other than interneurons. So the bottom clusters here in Tasic are interneurons, and the top ones are something else, but still inhibitory. The top-other has the Lamp cluster and the bottom-other has the Vis clusters.

Sst, Pvalb, Vip, and Lamp clusters are all inhibitory (inter)neurons. "Interneuron" simply means that the neuron does not project far away and all these inhibitory neurons are like that. That's exactly my confusion about the difference between 15 and 27 in Cao et al.

From your figures it's clear that 27 mostly maps to Vip+Lamp5 interneurons, while 15 is more all over the place but if anything also rather maps to the same Vip+Lamp5 area. Thanks!

dkobak:
Can you make it version 0.4.4? We are preparing a paper right now, and we'd like to refer to the exact version we are using. Or were you planning to do some other changes before releasing the version?

pavlin-policar:
Yeah, sure I can make a quick mini-release in the coming days. We'll get the other open PRs in at a later time.

PR #134, opened 2020-05-17
Pynndescent fixes

### LINKED ISSUE #130: Pynndescent build/query
We discussed this before, but I've been playing around with some sparse data now and wanted to report some runtimes.

When using pynndecent, openTSNE runs `build()` with n_neighbors=15 and then `query()` with n_neighbors=3*perplexity. At the same time, Leland said that that's not efficient and the recommended way to use pynndescent is to run `build()` with the desired number of neighbors and then simply take its constructed kNN graph without querying. You said that you ran some benchmarks and found your way to be faster. Here are runtimes I got on `X` that is sparse of size `(100000, 9630)`.

```
nn = NNDescent(X, metric='cosine', n_neighbors=15)      # Wall time: 39 s
nn.query(X, k=15)                                         # Wall time: 1min 57s
nn.query(X, k=90)                                         # Wall time: 3min 21s
nn90 = NNDescent(X, metric='cosine', n_neighbors=90)  # Wall time: 7min 45s
nn90.query(X, k=90)                                     # Wall time: 57min 53s
```

For k=90 it is indeed faster to build with k=15 and then query with k=90, so I can confirm your observation. 

My only suggestion would be to modify the `NNDescent` class so that if the desired `k` is less than some threshold then build is done with `k+1` and then the constructed tree is returned without `query`. We can simply use 15 as the threshold. I did this locally and can PR.

dkobak:
Fixes #130.

Changes:
1. Query() is only used for k>15.
2. n_jobs fixed to 1 for sparse inputs to avoid a pynndescent bug
3. find all points where index contains -1 values, and let them randomly attract each other.

### DIFF
--- a/openTSNE/nearest_neighbors.py
+++ b/openTSNE/nearest_neighbors.py
@@ -369,11 +369,25 @@ class NNDescent(KNNIndex):
         # unless we're actually going to use it
         import pynndescent
 
+        # Will use query() only for k>15
+        if k <= 15:
+            n_neighbors_build = k + 1
+        else:
+            n_neighbors_build = 15
+
+        # due to a bug, pynndescent currently does not support n_jobs>1
+        # for sparse inputs. This should be removed once it's fixed.
+        n_jobs_pynndescent = self.n_jobs
+        import scipy.sparse as sp
+
+        if sp.issparse(data):
+            n_jobs_pynndescent = 1
+
         # UMAP uses the "alternative" algorithm, but that sometimes causes
         # memory corruption, so use the standard one, which seems to work fine
         self.index = pynndescent.NNDescent(
             data,
-            n_neighbors=15,
+            n_neighbors=n_neighbors_build,
             metric=self.metric,
             metric_kwds=self.metric_params,
             random_state=self.random_state,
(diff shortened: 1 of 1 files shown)


dkobak:
I have no idea why this PR shows so many commits. Only the last commit is relevant. The rest is from my history, but the branch was synched to my master which was synched to your master. I must have messed something up with git. I can never really understand git :-/

pavlin-policar:
I think if you pull down the latest master and rebase the branch, it should get rid of all these commits. If you aren't used to rebasing, I would suggest backing up the changes to a file that git doesn't track.
```
git checkout master
git pull upstream master (assuming you have this repo set as upstream)
git checkout pynndescent_fixes
git rebase master
```
I suspect you didn't sync your master with the master branch on this repo, and that's why it pulled in that many commits. I have no idea what the merges are from though.

I'll go over the code tomorrow, but it would be really nice if you could add some tests to make sure it does work as expected.

dkobak:
Yeah, my master branch says that it's 17 commits ahead of yours, but this is nonsense because I pulled everything from your repository (upstream) and pushed to my master (origin). So they are the same. I don't understand why it would be 17 commits ahead.

I'll try the rebase command.

dkobak:
So actually in my case all four commands did nothing. 

```
$ git checkout master
Switched to branch 'master'
Your branch is up to date with 'origin/master'.
$ git pull upstream master
From https://github.com/pavlin-policar/openTSNE
 * branch            master     -> FETCH_HEAD
Already up to date.
$ git checkout pynndescent_fixes 
Switched to branch 'pynndescent_fixes'
$ git rebase master
Current branch pynndescent_fixes is up to date.
```

pavlin-policar:
Hmm, yeah if it says it's ahead by 17 commits, that's very strange. There are a couple of ways you can fix this, but the easiest is probably an interactive rebase, and you can fixup -- squash commits and discard commit messages -- all the commits except the last one. Or maybe even easier: copy over the changed file somewhere, delete the branch, checkout a new branch with the same name, copy over your changes, commit, and force push. Either way, it's a good idea to backup changes when doing any kind of git-gymnastics :)

pavlin-policar:
I should have been more specific -- you only needed to delete the branch locally, deleting the branch on Github closes the PR. Then if you force pushed the new branch, it would just override the older branch. It's no big deal, git takes a bit of trial and error, but once you've got it, it's pretty nice to work with. It's not letting me re-open this PR for some reason even though it's saying the branch has been restored. Just open another PR if you can't reopen it. No sense wasting time with this.

dkobak:
I tried that it but somehow it did not work. Whatever, I resolved it in my master, made a new branch and  opened a new PR.

PR #135, opened 2020-05-17
Workaround for -1 in pynndescent index

### LINKED ISSUE #130: Pynndescent build/query

dkobak:
Fixes #130 .

Changes:

*    Query() is only used for k>15.
*    n_jobs fixed to 1 for sparse inputs to avoid a pynndescent bug
*    find all points where index contains -1 values, and let them randomly attract each other.

### DIFF
--- a/openTSNE/nearest_neighbors.py
+++ b/openTSNE/nearest_neighbors.py
@@ -335,7 +335,7 @@ class NNDescent(KNNIndex):
             )
 
         if callable(metric):
-            from numba.targets.registry import CPUDispatcher
+            from numba.core.registry import CPUDispatcher
 
             if not isinstance(metric, CPUDispatcher):
                 warnings.warn(
--- a/openTSNE/nearest_neighbors.py
+++ b/openTSNE/nearest_neighbors.py
@@ -3,6 +3,7 @@ import warnings
 import numpy as np
 from scipy.spatial.distance import cdist
 from sklearn import neighbors
+from sklearn.utils import check_random_state
 
 from openTSNE import utils
 
@@ -430,7 +431,8 @@ class NNDescent(KNNIndex):
                     f"Run with verbose=True, to see indices of the failed points."
                 )
             distances[mask] = 1
-            fake_indices = np.random.choice(
+            rs = check_random_state(self.random_state)
+            fake_indices = rs.choice(
                 np.sum(mask), size=np.sum(mask) * indices.shape[1], replace=True
             )
             fake_indices = np.where(mask)[0][fake_indices]
--- a/openTSNE/nearest_neighbors.py
+++ b/openTSNE/nearest_neighbors.py
@@ -369,11 +369,25 @@ class NNDescent(KNNIndex):
         # unless we're actually going to use it
         import pynndescent
 
+        # Will use query() only for k>15
+        if k <= 15:
+            n_neighbors_build = k + 1
+        else:
+            n_neighbors_build = 15
+
+        # due to a bug, pynndescent currently does not support n_jobs>1
+        # for sparse inputs. This should be removed once it's fixed.
+        n_jobs_pynndescent = self.n_jobs
+        import scipy.sparse as sp
+
+        if sp.issparse(data):
+            n_jobs_pynndescent = 1
+
         # UMAP uses the "alternative" algorithm, but that sometimes causes
         # memory corruption, so use the standard one, which seems to work fine
         self.index = pynndescent.NNDescent(
             data,
-            n_neighbors=15,
+            n_neighbors=n_neighbors_build,
             metric=self.metric,
             metric_kwds=self.metric_params,
             random_state=self.random_state,
@@ -381,10 +395,29 @@ class NNDescent(KNNIndex):
             n_iters=n_iters,
             algorithm="standard",
             max_candidates=60,
-            n_jobs=self.n_jobs,
+            n_jobs=n_jobs_pynndescent,
         )
 
-        indices, distances = self.index.query(data, k=k + 1)
+        # -1 in indices means that pynndescent failed
+        indices, distances = self.index.neighbor_graph
+        mask = np.sum(indices == -1, axis=1) > 0
+
+        if k > 15:
+            indices, distances = self.index.query(data, k=k + 1)
+
+        # as a workaround, we let the failed points group together
+        if np.sum(mask) > 0:
+            warnings.warn(
+                f"`pynndescent` failed to find neighbors for some of the points."
+                f"As a workaround, openTSNE considers all such points similar to "
+                f"each other, so they will likely form a cluster in the embedding."
+            )
+            distances[mask] = 1
+            fake_indices = np.random.choice(
+                np.sum(mask), size=np.sum(mask) * indices.shape[1]
+            )
+            fake_indices = np.where(mask)[0][fake_indices]
+            indices[mask] = np.reshape(fake_indices, (np.sum(mask), indices.shape[1]))
 
         timer.__exit__()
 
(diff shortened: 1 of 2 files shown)


openTSNE/nearest_neighbors.py
```diff
@@ -369,22 +369,55 @@ def build(self, data, k):
         # unless we're actually going to use it
         import pynndescent
 
+        # Will use query() only for k>15
+        if k <= 15:
+            n_neighbors_build = k + 1
+        else:
+            n_neighbors_build = 15
+
+        # due to a bug, pynndescent currently does not support n_jobs>1
```
pavlin-policar, line 378: Capitalize first word of sentence.

openTSNE/nearest_neighbors.py
```diff
@@ -369,22 +369,55 @@ def build(self, data, k):
...
+        # for sparse inputs. This should be removed once it's fixed.
+        n_jobs_pynndescent = self.n_jobs
+        import scipy.sparse as sp
+
+        if sp.issparse(data):
```
pavlin-policar, line 383: This should trigger a warning.

openTSNE/nearest_neighbors.py
```diff
@@ -369,22 +369,55 @@ def build(self, data, k):
...
+        if k > 15:
+            indices, distances = self.index.query(data, k=k + 1)
+
+        # as a workaround, we let the failed points group together
+        if np.sum(mask) > 0:
+            warnings.warn(
+                f"`pynndescent` failed to find neighbors for some of the points."
+                f"As a workaround, openTSNE considers all such points similar to "
+                f"each other, so they will likely form a cluster in the embedding."
+            )
```
pavlin-policar, line 414: As discussed, I think there should be an easy way to get the indices of the failed points. Maybe this should be included in the warning only if `verbose=True` since you're getting a huge number of points failing, but it should definitely be there.

openTSNE/nearest_neighbors.py
```diff
@@ -369,22 +369,55 @@ def build(self, data, k):
...
+            distances[mask] = 1
+            fake_indices = np.random.choice(
+                np.sum(mask), size=np.sum(mask) * indices.shape[1]
```
pavlin-policar, line 417: I checked that the default for `np.random.choice` is `replace=True`, so this will work even if there aren't enough failed points. However, I still had to check since I don't know this by heart. Setting `replace=True` explicitly would be clearer and make it future proof, in case numpy ever decides to change their defaults.

dkobak:
Fixed everything as requested.

I was actually going to write that I don't think I will be able to look into the tests in the next days or even weeks, and might need to come back to it afterwards. Have some deadlines upcoming. If we can postpone the tests until later, or if you can add some, it would be great.

pavlin-policar:
Alright, I'll try to add some in the coming days.

dkobak:
Hey Pavlin, should we try to continue with this PR?

pavlin-policar:
Yeah, so here, I think three tests would be enough
1. make sure that pynndescent is actually run with `n_jobs=1` if input data is sparse
2. check if query is used if `k_neighbors` is more than 15
3. simulate a case where the indices are, in fact, -1 and make sure it works as intended
This will require a bit of mocking in the tests. Let me know if you want to tackle this.

dkobak:
To be honest, I am not sure how to address (1) and (2). What I could try to do, is to find some simple simulated case where pynndescent yields some -1 values.

pavlin-policar:
It should actually be fairly simple, it's just a case of mocking a few things and making sure things are called properly. It's no problem, I can probably get this done fairly quickly.

dkobak:
OK great. Meanwhile I discovered that I cannot install the latest pynndescent version (0.4.8) from conda due to some package conflicts. I had to reinstall my operating system and conda recently, so my conda environment is quite fresh. Weird. The latest UMAP installs without problem.  Can you install pynndescent=0.4.8?

pavlin-policar:
> Can you install pynndescent=0.4.8?

Yeah, I have no issues installing it. I tried in a fresh conda environment and installed it via `pip`. Also works with conda for me...

dkobak:
I think I localized the conflict: https://github.com/lmcinnes/pynndescent/issues/102

openTSNE/nearest_neighbors.py
```diff
@@ -371,7 +372,7 @@ def build(self, data, k):
 
         # Will use query() only for k>15
         if k <= 15:
-            n_neighbors_build = k + 1
+            n_neighbors_build = k
```
pavlin-policar, line 375: Before, when we used `query` everywhere, that also returned the point itself as a nearest neighbor. When building the index the "proper" way, this is not the case, so I changed this to just `k`. It's still works fine for query later on.

pavlin-policar:
@dkobak Was there anything else with this PR. I've added the tests, so if that's it, I'll merge. The tests are currently failing because numba removed something in the meantime, but that's fixed in the other PR.

openTSNE/nearest_neighbors.py
dkobak: Hmm. Are you sure it's not the case? You still do `return indices[:, 1:], distances[:, 1:]` later on, so will `index.neighbor_graph` contain 16 elements when you build with `n_neighbors=15`?

dkobak:
Wow, the tests look great. Never encountered `MagicMock` :-) I need to figure out how this stuff works. Thanks for taking care of it.

See my little comment above regarding k/k+1. Apart from that, I'm happy if you merge.

openTSNE/nearest_neighbors.py
pavlin-policar: Yes, you were right. At first I thought I should just fix it to return `indices, distances` as is, but then I checked and the nearst neighbor graph actually returns the points themselves as their closest neighbors with distance 0. That's quite unexpected, but I've fixed it now.

pavlin-policar:
Mocking things generally isn't the best way of testing things, but sometimes you really want to make sure some functions are called the right way, which is exactly what we need here.

Have you had a chance to look at #138 yet?

openTSNE/nearest_neighbors.py
dkobak: ok

dkobak:
Hmm, after we implemented this workaround, I realized that the last version of pynndescent does not return any -1s anymore in the datasets I have. I asked here https://github.com/lmcinnes/pynndescent/issues/99#issuecomment-654128474 and Leland said that there were some changes in how cosine distance is handled that made it less likely for pynndescent to fail to find neighbors. Oh well.

pavlin-policar:
That's fine. In case it ever pops up again, we have a solution, and we wouldn't get strange errors.

PR #143, opened 2020-10-17
Add Python 3.9 support

pavlin-policar:
##### Description of changes
Add Python 3.9 to tests and release pipelines.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/azure-pipelines-release.yml
+++ b/azure-pipelines-release.yml
@@ -28,6 +28,9 @@ jobs:
         osx - python38:
           image.name: 'macos-10.14'
           python.version: '3.8'
+        osx - python39:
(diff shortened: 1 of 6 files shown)


pavlin-policar:
Premature -- python 3.9 is not yet available on Azure pipelines windows and mac image.

pavlin-policar:
The Windows VM now contains Python 3.9. So this is no longer blocked.

However, the problem now is numba, as Python 3.9 is not yet supported: https://github.com/numba/numba/issues/6345

However, since pynndescent is an optional dependency, I don't think it makes sense to hold up the release because of this.

PR #144, opened 2020-10-17
Update to annoy v1.17.0

pavlin-policar:
##### Description of changes
Update annoy dependency to v1.17. This should be faster than before, as index building now uses multiple cores. https://github.com/spotify/annoy/pull/495

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/openTSNE/dependencies/annoy/annoylib.h
+++ b/openTSNE/dependencies/annoy/annoylib.h
@@ -58,6 +58,12 @@ typedef signed __int64    int64_t;
 #include <queue>
 #include <limits>
 
+#ifdef ANNOYLIB_MULTITHREADED_BUILD
+#include <thread>
+#include <mutex>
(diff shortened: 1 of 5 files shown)


dkobak:
What's the reason not to depend on annoy package, now that it has wheels and whatnot? Or does it not?

pavlin-policar:
Oh, I wasn't aware they had made wheels. However, after looking at their [PyPI](https://pypi.org/project/annoy/#files), they only have a wheel for a single version of OSX and a single version of Python. If anything, Windows is problematic because it doesn't come with a compiler by default and there are no wheels for Windows unfortunately.

dkobak:
I see. Wasn't aware of that. I only saw that you had some exchange with annoy developers over there a few months ago, and my impression back then was that everything was implemented. But I guess not.

Issue #145, opened 2020-10-19
Pass verbose=1 to pynndescent?


dkobak:
When run with `verbose=1`, pynndescent provides some helpful verbose output:

```
Mon Oct 19 12:44:31 2020 Building RP forest with 55 trees
Mon Oct 19 12:50:16 2020 metric NN descent for 20 iterations
	 1  /  20
	 2  /  20
	 3  /  20
	 4  /  20
	 5  /  20
	Stopping threshold met -- exiting after 5 iterations
Mon Oct 19 12:54:43 2020 Finished Nearest Neighbor Search
```

`openTSNE` currently does not pass the `verbose` setting into `pynndescent`, so the output with `verbose=1` looks like this:

```
===> Finding 15 nearest neighbors using NN descent approximate search using cosine distance...
   --> Time elapsed: 980.70 seconds
```

Would it make sense to pass the `verbose` flag? The logging format used by `pynndescent` looks a bit different, but I think it does not matter that much, and verbose output could be useful.

pavlin-policar:
Maybe we could do this with verbosity levels. E.g. with `verbose=True` i.e. `verbose=1` I would keep the output as is because I don't think I'd want to look at that output every time. But then we could do `verbose=2` to show that as well. What do you think?

dkobak:
Makes sense!

PR #142, opened 2020-10-17
Fixes #139

pavlin-policar:
Deprecate neighbors_method attribute and replace it with neighbors to
satisfy scikit-learn warnings in repr.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

PR #147, opened 2020-10-20
pynndescent: n_jobs, metrics, verbose

### LINKED ISSUE #146: Pynndescent in t-SNE with k=15 seems a lot slower than in UMAP
I've just noticed that UMAP 0.4.6 with pynndescent 0.5 can be quite a bit faster to construct the kNN graph compared to openTSNE with uniform kernel and k=15. This does not make sense to me. I added verbose=1 to the pynndescent call within openTSNE and here is what I see:

UMAP - 34 sec:

```
UMAP(angular_rp_forest=True, metric='cosine', verbose=1)
Construct fuzzy simplicial set
Mon Oct 19 13:54:22 2020 Finding Nearest Neighbors
Mon Oct 19 13:54:22 2020 Building RP forest with 21 trees
Mon Oct 19 13:54:32 2020 metric NN descent for 17 iterations
	 1  /  17
	 2  /  17
	 3  /  17
	 4  /  17
	 5  /  17
	 6  /  17
	Stopping threshold met -- exiting after 6 iterations
Mon Oct 19 13:54:58 2020 Finished Nearest Neighbor Search
```

tSNE -- 61 sec

```
===> Finding 15 nearest neighbors using NN descent approximate search using cosine distance...
Mon Oct 19 14:03:15 2020 Building RP forest with 21 trees
Mon Oct 19 14:03:33 2020 metric NN descent for 17 iterations
	 1  /  17
	 2  /  17
	 3  /  17
	 4  /  17
	 5  /  17
	 6  /  17
	Stopping threshold met -- exiting after 6 iterations
   --> Time elapsed: 61.53 seconds
```

I ran both snippets twice to make sure all numba code was pre-compiled.

I don't know why this happens. The `pynndescent` call in UMAP looks identical to the one in t-SNE. https://github.com/lmcinnes/umap/blob/master/umap/umap_.py#L290

dkobak:
Fixes #146, #145, and updates metrics as per #142.

### DIFF
--- a/openTSNE/nearest_neighbors.py
+++ b/openTSNE/nearest_neighbors.py
@@ -390,7 +390,7 @@ class NNDescent(KNNIndex):
             n_iters=n_iters,
             max_candidates=60,
             n_jobs=self.n_jobs,
-            verbose=1 if self.verbose>1 else 0,
+            verbose=self.verbose > 1,
         )
 
         # -1 in indices means that pynndescent failed
--- a/openTSNE/nearest_neighbors.py
+++ b/openTSNE/nearest_neighbors.py
@@ -277,6 +277,7 @@ class NNDescent(KNNIndex):
         # general minkowski distances
         "euclidean",
         "l2",
+        "sqeuclidean",
         "manhattan",
         "taxicab",
         "l1",
@@ -294,11 +295,14 @@ class NNDescent(KNNIndex):
         # Other distances
         "canberra",
         "cosine",
+        "dot",
         "correlation",
         "hellinger",
         "haversine",
         "braycurtis",
         "spearmanr",
+        "kantorovich",
+        "wasserstein",
         # Binary distances
         "hamming",
         "jaccard",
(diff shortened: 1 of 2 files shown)


dkobak:
I don't understand why all these tests failed :-(

pavlin-policar:
[This test](https://github.com/pavlin-policar/openTSNE/blob/master/tests/test_nearest_neighbors.py#L279-L283) will need to be updated :)

openTSNE/nearest_neighbors.py
```diff
@@ -398,7 +389,8 @@ def build(self, data, k):
             n_trees=n_trees,
             n_iters=n_iters,
             max_candidates=60,
-            n_jobs=n_jobs_pynndescent,
+            n_jobs=self.n_jobs,
+            verbose=1 if self.verbose>1 else 0,
```
pavlin-policar, line 393: A bit nit-picky, but could you add spaces around the `>`, so `verbose=1 if self.verbose > 1 else 0`, or better yet `verobose=self.verbose > 1`.

dkobak: Yes, that's a better way to write it, thanks. Done.

dkobak:
Oh right. I fixed the test. I'm always confused when one of the tests fails because it's not obvious which test has failed...

pavlin-policar:
Thanks! Also, if you scroll to the top of the [test output](https://dev.azure.com/pavlingp/openTSNE/_build/results?buildId=304&view=logs&j=9900b494-448c-5226-d8d0-3aa940fadb03&t=28aaf5cc-09e3-5d10-107a-1e82bc77f3c5), you can see exactly which tests failed. E.g. line 118.

PR #148, opened 2020-10-25
Add hnswlib as k nearest neighbour index

TimRepke:
##### Issue
The kNN algorithms available in openTSNE are not as fast and/or memory efficient as HNSW (https://arxiv.org/abs/1603.09320).
Furthermore, Annoy, BallTree, and pynndescent are not made for high-dimensional data; For example, Annoy even throws an error above 1000 dimensions.

##### Description of changes
This pull request ...
* adds hnswlib as a dependency
* adds option to choose 'hnswlib' to `build_knn_index` for `Affinities`
* implements additional `KNNIndex` class to interface with hnswlib

##### Includes
- [X] Code changes
- [x] Tests
- [x] Documentation

If you like the idea of having HNSW as an alternative, I'll add tests and some documentation.

### DIFF
--- a/openTSNE/nearest_neighbors.py
+++ b/openTSNE/nearest_neighbors.py
@@ -12,7 +12,7 @@ class KNNIndex:
     VALID_METRICS = []
 
     def __init__(
-            self, metric, metric_params=None, n_jobs=1, random_state=None, verbose=False
+        self, metric, metric_params=None, n_jobs=1, random_state=None, verbose=False
     ):
         self.index = None
         self.metric = self.check_metric(metric)
@@ -94,7 +94,7 @@ class BallTree(KNNIndex):
             effective_metric = "euclidean"
             effective_data = data.copy()
             effective_data = (
-                    effective_data / np.linalg.norm(effective_data, axis=1)[:, None]
+                effective_data / np.linalg.norm(effective_data, axis=1)[:, None]
             )
             # In order to properly compute cosine distances when querying the
             # index, we need to store the original data
@@ -140,7 +140,7 @@ class BallTree(KNNIndex):
         if self.metric == "cosine":
             effective_data = query.copy()
             effective_data = (
-                    effective_data / np.linalg.norm(effective_data, axis=1)[:, None]
+                effective_data / np.linalg.norm(effective_data, axis=1)[:, None]
             )
         else:
             effective_data = query
@@ -330,7 +330,7 @@ class NNDescent(KNNIndex):
         import pynndescent
 
         if not np.array_equal(
-                list(pynndescent.distances.named_distances), self.VALID_METRICS
+            list(pynndescent.distances.named_distances), self.VALID_METRICS
         ):
             warnings.warn(
                 "`pynndescent` has recently changed which distance metrics are supported, "
--- a/openTSNE/nearest_neighbors.py
+++ b/openTSNE/nearest_neighbors.py
@@ -12,7 +12,7 @@ class KNNIndex:
     VALID_METRICS = []
 
     def __init__(
-        self, metric, metric_params=None, n_jobs=1, random_state=None, verbose=False
+            self, metric, metric_params=None, n_jobs=1, random_state=None, verbose=False
     ):
         self.index = None
         self.metric = self.check_metric(metric)
@@ -94,7 +94,7 @@ class BallTree(KNNIndex):
             effective_metric = "euclidean"
             effective_data = data.copy()
             effective_data = (
-                effective_data / np.linalg.norm(effective_data, axis=1)[:, None]
+                    effective_data / np.linalg.norm(effective_data, axis=1)[:, None]
             )
             # In order to properly compute cosine distances when querying the
             # index, we need to store the original data
@@ -140,7 +140,7 @@ class BallTree(KNNIndex):
         if self.metric == "cosine":
             effective_data = query.copy()
             effective_data = (
-                effective_data / np.linalg.norm(effective_data, axis=1)[:, None]
+                    effective_data / np.linalg.norm(effective_data, axis=1)[:, None]
             )
         else:
             effective_data = query
@@ -330,7 +330,7 @@ class NNDescent(KNNIndex):
         import pynndescent
 
         if not np.array_equal(
-            list(pynndescent.distances.named_distances), self.VALID_METRICS
+                list(pynndescent.distances.named_distances), self.VALID_METRICS
         ):
             warnings.warn(
                 "`pynndescent` has recently changed which distance metrics are supported, "
@@ -517,7 +517,7 @@ class HNSW(KNNIndex):
         timer.__enter__()
 
         # Set ef parameter for (ideal) precision/recall
-        self.index.set_ef(2 * k)
+        self.index.set_ef(min(2 * k, self.index.get_current_count()))
 
         # Query for kNN
         indices, distances = self.index.knn_query(query, k=k, num_threads=self.n_jobs)
--- a/openTSNE/nearest_neighbors.py
+++ b/openTSNE/nearest_neighbors.py
@@ -12,7 +12,7 @@ class KNNIndex:
     VALID_METRICS = []
 
     def __init__(
-            self, metric, metric_params=None, n_jobs=1, random_state=None, verbose=False
+        self, metric, metric_params=None, n_jobs=1, random_state=None, verbose=False
     ):
         self.index = None
         self.metric = self.check_metric(metric)
@@ -94,7 +94,7 @@ class BallTree(KNNIndex):
             effective_metric = "euclidean"
             effective_data = data.copy()
             effective_data = (
-                    effective_data / np.linalg.norm(effective_data, axis=1)[:, None]
+                effective_data / np.linalg.norm(effective_data, axis=1)[:, None]
             )
             # In order to properly compute cosine distances when querying the
             # index, we need to store the original data
@@ -140,7 +140,7 @@ class BallTree(KNNIndex):
         if self.metric == "cosine":
             effective_data = query.copy()
             effective_data = (
-                    effective_data / np.linalg.norm(effective_data, axis=1)[:, None]
+                effective_data / np.linalg.norm(effective_data, axis=1)[:, None]
             )
         else:
             effective_data = query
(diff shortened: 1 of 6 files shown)


openTSNE/nearest_neighbors.py
```diff
@@ -445,3 +445,77 @@ def query(self, query, k):
...
+        )
+        timer.__enter__()
+
+        from hnswlib import Index
+
+        hnsw_space = {
+            "cosine": "cosine",
+            "dot": "ip",
+            "euclidean": "l2"
+        }[self.metric]
```
TimRepke, line 472: If this should be merged, remember to add
```
"l2": "l2",
"ip": "ip"
```
to the alias dictionary.

pavlin-policar:
Thank you for this PR, I think it would be a great idea to support `hnswlib`, and it's something I have looked at in the past. 

However, as it stands, we can't add a hard dependency onto `hnswlib`. While `hnswlib` is available on [PyPI](https://pypi.org/project/hnswlib/#files), there's only a source distribution. Because `hnswlib` is a C++ library, this will need to be compiled on the users machine, which is not okay. This is especially problematic on Windows machines, which do not come with a C++ compiler by default. What we would need are precompiled wheels on all major platforms for different Python versions. Furthermore, `openTSNE` is also available on conda-forge, so there would also need to be a conda-forge package of `hnswlib`, which there currently isn't. So a hard dependency is definitely a no go.

I would be very happy to add it as an optional dependency though. So if you wanted to use `hnswlib` with `openTSNE`, you would just manually install it into your environment. This is what we currently do with `pynndescent`. The changes to this PR should be fairly minimal, just take a look at what we do for [`NNDescent`](https://github.com/pavlin-policar/openTSNE/blob/master/openTSNE/nearest_neighbors.py#L275-L446). And also, please add unit tests, something like [`TestAnnoy`](https://github.com/pavlin-policar/openTSNE/blob/master/tests/test_nearest_neighbors.py#L74-L75) should be fine.

TimRepke:
Thank you for the feedback. I didn't have the compatibility issues in mind, but adding a hard dependency felt wrong somehow anyway.

Unfortunately, I won't really have time before Wednesday to make this PR nice (including proper tests and documentation). But I set a reminder for myself, so you can expect an update next week for sure!

pavlin-policar:
> Unfortunately, I won't really have time before Wednesday to make this PR nice (including proper tests and documentation). But I set a reminder for myself, so you can expect an update next week for sure!

That's not a problem at all. I appreciate the PR! I would also be interested in seeing benchmarks between annoy and hnswlib. I was under the impression that annoy was pretty fast.

TimRepke:
Hi,

I updated the pull request
* Mark hnswlib as an optional dependency (also did this for pynndescend)
* Add hint for hnsw index to docstrings
* Add sensible import check
* Add pip install commands to azure setups
* Update metric aliases for hnsw (originally, ip and l2 were missing)
* Update metric_param handling (originally, missing arguments might have triggered failure

Notes
* I did not change the auto-index setup (I guess that's fine as is)
* I added tests for exact nearest neighbour matches, those fail (it's an approximate NN index afterall...); not included here
  * Should I add them back in? There are 25/1500 (1.67%) mismatched elements. There could be a threshold check.
* As far as I can tell, hnswlib has no true random initialisation, they default their random seed if not provided. Is that an issue?

openTSNE/nearest_neighbors.py
```diff
@@ -445,3 +445,89 @@ def query(self, query, k):
...
+            "cosine": "cosine",
+            "dot": "ip",
+            "euclidean": "l2",
+            "ip": "ip",
+            "l2": "l2"
+        }[self.metric]
+
+        self.index = Index(space=hnsw_space, dim=data.shape[1])
+
+        metric_params = {
```
TimRepke, line 488: @pavlin-policar Is this the intended use of `metric_params`? It does work, but the naming is a little bit confusing. Here, it is more used as `index_params`.

openTSNE/nearest_neighbors.py
```diff
@@ -445,3 +445,89 @@ def query(self, query, k):
...
+    def query(self, query, k):
+        timer = utils.Timer(
+            f"Finding {k} nearest neighbors in existing embedding using HNSWlib "
+            f"approximate search...",
+            self.verbose,
+        )
+        timer.__enter__()
+
+        # Set ef parameter for (ideal) precision/recall
+        self.index.set_ef(2 * k)
```
TimRepke, line 524: This is a fixed setting I typically use. `hnswlib` recommends any value within `k < ef < index size`.  
https://github.com/nmslib/hnswlib/blob/master/ALGO_PARAMS.md

openTSNE/nearest_neighbors.py
pavlin-policar: No, you're using `metric_params` to pass parameters to the index construction. The idea of metric params is in case the metric we use takes additional params, as indicated by [`scikit-learn`](https://scikit-learn.org/stable/modules/generated/sklearn.neighbors.NearestNeighbors.html#sklearn.neighbors.NearestNeighbors). I don't think I've ever had to use this.

We currently don't have any way to actually alter the index parameters, as you're doing here. Which kind of takes away from the flexibility I guess. But that kind of forces us to have good defaults.

Please change this though, since this is not the intedend usage of `metric_params.`

openTSNE/nearest_neighbors.py
```diff
@@ -12,7 +12,7 @@ class KNNIndex:
     VALID_METRICS = []
 
     def __init__(
-        self, metric, metric_params=None, n_jobs=1, random_state=None, verbose=False
+            self, metric, metric_params=None, n_jobs=1, random_state=None, verbose=False
```
pavlin-policar, line 15: Please remove the added spacing here.

openTSNE/nearest_neighbors.py
```diff
@@ -94,7 +94,7 @@ def build(self, data, k):
             effective_metric = "euclidean"
             effective_data = data.copy()
             effective_data = (
-                effective_data / np.linalg.norm(effective_data, axis=1)[:, None]
+                    effective_data / np.linalg.norm(effective_data, axis=1)[:, None]
```
pavlin-policar, line 97: and here

openTSNE/nearest_neighbors.py
```diff
@@ -140,7 +140,7 @@ def query(self, query, k):
         if self.metric == "cosine":
             effective_data = query.copy()
             effective_data = (
-                effective_data / np.linalg.norm(effective_data, axis=1)[:, None]
+                    effective_data / np.linalg.norm(effective_data, axis=1)[:, None]
```
pavlin-policar, line 143: and here

openTSNE/nearest_neighbors.py
```diff
@@ -330,7 +330,7 @@ def check_metric(self, metric):
         import pynndescent
 
         if not np.array_equal(
-            list(pynndescent.distances.named_distances), self.VALID_METRICS
+                list(pynndescent.distances.named_distances), self.VALID_METRICS
```
pavlin-policar, line 333: and here

openTSNE/nearest_neighbors.py
pavlin-policar: Hmm, I don't know, I think we definitely need this check. Using high perplexity values is useful in a lot of cases, so we need a lot of neighbors. So we definitely can't have this triggering an error. I would allow any k, actually. Why would we limit the number to (2*k). In that case it wouldn't make sense to use HNSW at all, since we'd need the entire distance matrix. But I'd leave that up to the user. Maybe a warning...

setup.py
```diff
@@ -306,9 +306,12 @@ def readme():
     install_requires=[
         "numpy>=1.14.6",
         "scikit-learn>=0.20",
-        "scipy",
+        "scipy"
```
pavlin-policar, line 309: please don't remove trailing commas

pavlin-policar (changes requested):
Sorry, this took me so long to get to. Overall, I think it looks really good. I also ran it locally on a 44k data set, and it took 2 seconds! Annoy took 10 seconds. So that's really quite impressive.

I have a few nitpicks regarding code style, and I'd appreciate if you could fix them. Mainly, you changed a few lines where there are no changes by introducing white-space. This will mess up the git history, so remove that. I also generally use trailing commas, but that's really a nitpick.

If you could change the `metric_params` and the `k` check that you mentioned, I think this is fine to merge.

pavlin-policar:
I'm probably going to have to delay the next release until numba supports py39, which is supposedly going to be by mid-december. So if you can finalize this PR in the next two weeks, that would be amazing. I'd be really happy to include this in the next release!

openTSNE/nearest_neighbors.py
TimRepke: I checked it, `hnswlib` doesn't throw any errors. I limited the highest number that can be set anyway just to be sure nothing weird happens under the hood.

openTSNE/nearest_neighbors.py
TimRepke: Fair enough. I removed this part for now, it could stay in as an undocumented hack though :P
There seems to be no easy way to add the ability to configure the defaults from the outside, so I set the defaults from the HNSW examples.

openTSNE/nearest_neighbors.py
TimRepke: Sorry, kind of muscle memory to run PEP8 check once I wrote a few lines, so apparently my IDE reformatted other parts. I fixed that now.

TimRepke:
Hopefully captured all your points. Let me know what needs updating :-)
Do you prefer having squashed commits per PR? Not sure how well GitHub likes it across repositories, but I could try.

openTSNE/nearest_neighbors.py
pavlin-policar: The first commit you added fixed this, the second one re-added this spacing. I know this isn't technically PEP8 compliant, but we follow the black style guide. I think it's cleaner than strict adherence to PEP8.

pavlin-policar:
Perfect! Thanks for indulging my nitpicks. I squashed them here, because there were 3 commits called "Clean up code style" :) Thanks for all your help. I haven't done any proper benchmarks, but HNSW seems to be even faster than annoy.

TimRepke:
No worries, it's an honour to contribute :-)

Issue #156, opened 2021-01-29
Add optional k_neighbors input parameter to PerplexityBasedNN


dkobak:
Currently `PerplexityBasedNN` uses the standard heuristic 
```
k_neighbors = min(self.n_samples - 1, int(3 * self.perplexity))
```
What do you think about adding an input parameter `k_neighbors='auto'`. The default `'auto'` would use the current behaviour. But e.g. if `perplexity=30, k_neighbors=40`, then that's going to be used as is. Or also `perplexity=30, k_neighbors=200`.

This came up in a project I am doing with a rotation MSc student, so I was wondering if it would make sense to allow this flexibility to PerplexityBasedNN. I think it'd be nice because currently `3` in `3 * self.perplexity` is a hard-coded constant which is not great.

pavlin-policar:
I have thought about this in the context of scanpy, where this behaviour might be desirable, but it's not really clear to me how this would play with calculating the Gaussian kernel widths. Intuitively, I don't think it should really affect them much, because the perplexity of the distribution can be calculated on any number of samples. I would assume that just the accuracy would suffer if we used too few samples i.e. the kernel bandwidths might be estimated incorrectly. Not that I think this would be problematic, since the uniform kernel works just as well.

I think this is something we will have to add to accomodate the scanpy use-case, but we'll have to figure out what parameter value combinations are reasonable. 3\*k is hard-coded because it makes sense. Neighbors further than the 3*k nearest neighbors will be assigned a practically zero value, so parameter combinations like `perplexity=30, k=500` should be heavily discouraged. On the other hand, running something like `perplexity=500, k=30` would make no sense and would lead to incorrect results.

Off the top of my head, something like
```python
if k > 3 * perplexity:
    warn
elif k < perplexity * 2:
    error or at least warn
```

dkobak:
It's mathematically impossible to obtain perplexity larger than the number of neighbors. Uniform kernel yields maximum perplexity, and that's equal to the number of neighbors. So if `k > perplexity`, there should be an error. We can give a warning if `k > 3 * perplexity`, why not.

pavlin-policar:
Hmm, you're right, but I don't really understand why. I've played around with this a little, and, as you said, the moment `k < perplexity`, the bandwidths become basically zero. What's the intuition behind this?

But, basically, yes, that would definitely warrant an error!

dkobak:
I think the Gaussian variance sigma should go to infinity in this limit, not to zero.

[...]

PR #157, opened 2021-03-01
Allow to set k_neighors in perplexity based affinity

dkobak:
This should implement my suggestion in #156.

I checked and it seems to work as expected. 

However, notice that I changed the logic in `to_new` and in `set_perplexity` a little, to make it consistent with `k_neighbors` that could possibly be user-provided. For simplicity, I made these functions always use the same originally provided or computed `k_neighbors` neighbors. So e.g. if the original affinity is computed with `perplexity=20, k_neighbors=25`, then `to_new(perplexity=10)` will use the same `k_neighbors=25`. Previously your code would have used 3*10=30 in this case (which won't work in this particular situation because 25<30).

The same for `set_perplexity`. 

This may result in some loss of efficiency, but I thought it's probably negligible. Let me know what you think.

### DIFF
--- a/openTSNE/affinity.py
+++ b/openTSNE/affinity.py
@@ -138,21 +138,21 @@ class PerplexityBasedNN(Affinities):
         self.n_samples = data.shape[0]
         
         if k_neighbors=="auto":
-            self.k_neighbors = min(self.n_samples - 1, int(3 * perplexity))
+            self._k_neighbors = min(self.n_samples - 1, int(3 * perplexity))
         else:
-            self.k_neighbors = k_neighbors
+            self._k_neighbors = k_neighbors
 
         self.perplexity = self.check_perplexity(perplexity)
         self.verbose = verbose
         
-        if self.k_neighbors > int(3 * self.perplexity):
+        if self._k_neighbors > int(3 * self.perplexity):
             log.warning(
                 "The k_neighbors value is over 3 times larger than the perplexity value. "
                 "This may result in an unnecessary slowdown." 
             )
         
         self.knn_index, self.__neighbors, self.__distances = build_knn_index(
-            data, method, self.k_neighbors, metric, metric_params, n_jobs, 
+            data, method, self._k_neighbors, metric, metric_params, n_jobs, 
             random_state, verbose
         )
 
@@ -170,12 +170,13 @@ class PerplexityBasedNN(Affinities):
     def set_perplexity(self, new_perplexity):
         """Change the perplexity of the affinity matrix.
 
-        Note that we only allow lowering the perplexity or restoring it to its
-        original value. This restriction exists because setting a higher
-        perplexity value requires recomputing all the nearest neighbors, which
-        can take a long time. To avoid potential confusion as to why execution
-        time is slow, this is not allowed. If you would like to increase the
-        perplexity above the initial value, simply create a new instance.
+        Note that we only allow setting the perplexity to a value not larger
+        than the number of neighbors used for the original perplexity. This
+        restriction exists because setting a higher perplexity value requires
+        recomputing all the nearest neighbors, which can take a long time.
+        To avoid potential confusion as to why execution time is slow, this
+        is not allowed. If you would like to increase the perplexity above 
+        that value, simply create a new instance.
 
         Parameters
         ----------
@@ -189,7 +190,7 @@ class PerplexityBasedNN(Affinities):
         # Verify that the perplexity isn't too large
         new_perplexity = self.check_perplexity(new_perplexity)
         # Recompute the affinity matrix
-        if new_perplexity > self.perplexity:
+        if new_perplexity > self.__neighbors.shape[1]:
             raise RuntimeError(
                 "The desired perplexity `%.2f` is larger than the initial one "
                 "used. This would need to recompute the nearest neighbors, "
@@ -253,8 +254,12 @@ class PerplexityBasedNN(Affinities):
         """
         perplexity = perplexity if perplexity is not None else self.perplexity
         perplexity = self.check_perplexity(perplexity)
+         
+        _k_neighbors = int(3*perplexity)
+        if _k_neighbors > self._k_neighbors:
+            _k_neighbors = self._k_neighbors
 
-        neighbors, distances = self.knn_index.query(data, self.k_neighbors)
+        neighbors, distances = self.knn_index.query(data, _k_neighbors)
 
         with utils.Timer("Calculating affinity matrix...", self.verbose):
             P = joint_probabilities_nn(
@@ -276,8 +281,8 @@ class PerplexityBasedNN(Affinities):
         if perplexity <= 0:
             raise ValueError("Perplexity must be >=0. %.2f given" % perplexity)
 
-        if self.k_neighbors < perplexity:
-            old_perplexity, perplexity = perplexity, self.k_neighbors / 3
+        if self._k_neighbors < perplexity:
+            old_perplexity, perplexity = perplexity, self._k_neighbors / 3
             log.warning(
                 "Perplexity value %d is too high. Using perplexity %.2f instead"
                 % (old_perplexity, perplexity)
@@ -290,7 +295,7 @@ def build_knn_index(
     data, method, k, metric, metric_params=None, n_jobs=1, random_state=None, verbose=False
 ):
     preferred_approx_method = nearest_neighbors.Annoy
-    if is_package_installed("pynndescent") and sp.issparse(data) and metric not in [
+    if is_package_installed("pynndescent") and (sp.issparse(data) or metric not in [
         "cosine",
         "euclidean",
         "manhattan",
@@ -299,7 +304,7 @@ def build_knn_index(
         "l1",
         "l2",
         "taxicab",
-    ]:
+    ]):
         preferred_approx_method = nearest_neighbors.NNDescent
 
     if data.shape[0] < 1000:
--- a/openTSNE/affinity.py
+++ b/openTSNE/affinity.py
@@ -265,7 +265,7 @@ class PerplexityBasedNN(Affinities):
         perplexity = perplexity if perplexity is not None else self.perplexity
         perplexity = self.check_perplexity(perplexity, self.n_samples)
          
-        k_neighbors = min(self.n_samples, int(3 * new_perplexity))
+        k_neighbors = min(self.n_samples, int(3 * perplexity))
 
         neighbors, distances = self.knn_index.query(data, k_neighbors)
 
--- a/openTSNE/affinity.py
+++ b/openTSNE/affinity.py
@@ -138,21 +138,21 @@ class PerplexityBasedNN(Affinities):
         self.n_samples = data.shape[0]
         
         if k_neighbors=="auto":
-            self._k_neighbors = min(self.n_samples - 1, int(3 * perplexity))
+            _k_neighbors = min(self.n_samples - 1, int(3 * perplexity))
         else:
-            self._k_neighbors = k_neighbors
+            _k_neighbors = k_neighbors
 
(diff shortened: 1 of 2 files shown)


openTSNE/affinity.py
```diff
@@ -129,14 +133,27 @@ def __init__(
...
         random_state=None,
         verbose=False,
+        k_neighbors="auto",
     ):
         self.n_samples = data.shape[0]
+        
+        if k_neighbors=="auto":
+            self.k_neighbors = min(self.n_samples - 1, int(3 * perplexity))
+        else:
+            self.k_neighbors = k_neighbors
```
pavlin-policar, line 143: Please change this so `self.k_neighbors` is set to the user-defined parameter. So, if I instantiate `x = Class(k="auto")`, then I'd be surprised if `x.k` is set to `90` for instance. I'd expect it to be set to `"auto"`. This type of interface also breaks if I change `x.k` myself e.g. `x.k = "auto"`. Then, nothing will work anymore.

I realise it may not be as relevant here because it's all going on inside the same function, but it's good practice. I also realise now this is technically not what happens with `self.perplexity`, but at least there's a warning there.

So I would suggest saving `self.k_neighbors = "auto"`, and using a local variable to determine the proper number of neighbors. Then, we'll have to do something similar in `to_new`.

openTSNE/affinity.py
```diff
@@ -172,8 +189,7 @@ def set_perplexity(self, new_perplexity):
         # Verify that the perplexity isn't too large
         new_perplexity = self.check_perplexity(new_perplexity)
         # Recompute the affinity matrix
-        k_neighbors = min(self.n_samples - 1, int(3 * new_perplexity))
-        if k_neighbors > self.__neighbors.shape[1]:
+        if new_perplexity > self.perplexity:
```
pavlin-policar, line 192: I don't see the point of this change. As you implemented it now, we can choose a k that is larger than 3*perplexity. So in that case, if I instantiate `a = Aff(perplexity=30, k_neighbors=500)`, then I should be able to call `a.set_perplexity(100)` without any issues and use 300 neighbors. I should even be able to call `a.set_perplexity(500)`, since those are available.

openTSNE/affinity.py
```diff
@@ -237,9 +253,8 @@ def to_new(self, data, perplexity=None, return_distances=False):
         """
         perplexity = perplexity if perplexity is not None else self.perplexity
         perplexity = self.check_perplexity(perplexity)
-        k_neighbors = min(self.n_samples - 1, int(3 * perplexity))
 
-        neighbors, distances = self.knn_index.query(data, k_neighbors)
+        neighbors, distances = self.knn_index.query(data, self.k_neighbors)
```
pavlin-policar, line 257: I am not a fan of this at all. For instance, if I create an embedding using 100k points using perplexity 500, then I want to add 1mln data points into that embedding, this means that we'll be calculating 500 neighbors for each 1mln data points. The number of neighbors can be pretty low in the transform functionality for it to work well, so this would really slow down `.transform` in that use case at very little benefit.

I think `transform` should have its own set of parameters, both for perplexity and k_neighbors.

pavlin-policar (changes requested):
Thanks for tackling this issue! I have several comments regarding the implementation, but I think the general gist of it good.

Also, two tests seem to be failing, and it't not obvious to me why. They seem like things that should definitely still be passing.

openTSNE/affinity.py
dkobak: Fixed. I had the same issue raised with my recent PR to sklearn. They use `self._k_neighbors` naming scheme for this purpose, so that's what I used here now.

openTSNE/affinity.py
dkobak: Fair enough, fixed such that the error only happens whenever `new_perplexity > self.__neighbors.shape[1]`.

openTSNE/affinity.py
dkobak: Fair enough. Fixed so that `3*perplexity` is used in `to_new()` unless this is larger than `self._k_neighbors`. If so, `self._k_neighbors` is used.

dkobak:
I haven't looked at the tests yet -- will do after we iron out the main logic.

openTSNE/affinity.py
pavlin-policar: Hmm, I'm not sure I see the point of this. Why would we want to limit the number of nearest neighbors when embedding  new data points to the number of nearest neighbors we used when constructing the embedding? They seem like two independent things to me. Any KNN index should be able to be queried with an arbitrary number of neighbors. If I wanted to, I might want to build an embedding using perplexity=30, then use perplexity=500 when adding new data points. I don't see any reason we might want to disallow this.

I think the number of neighbors should be completely seprate between the initial `fit` and `transform` functions.

openTSNE/affinity.py
```diff
@@ -129,14 +133,27 @@ def __init__(
...
         random_state=None,
         verbose=False,
+        k_neighbors="auto",
     ):
         self.n_samples = data.shape[0]
+        
+        if k_neighbors=="auto":
+            self._k_neighbors = min(self.n_samples - 1, int(3 * perplexity))
+        else:
+            self._k_neighbors = k_neighbors
```
pavlin-policar, line 143: If we agree that the `k` parameter in `transform` should be completely independent of the number of nearest neighbors used to construct the embedding, I don't see a reason to store this on `self`. It just pollutes the class with unnecessary attributes. A local variable should be fine here.

openTSNE/affinity.py
```diff
@@ -188,8 +205,8 @@ def set_perplexity(self, new_perplexity):
             "Perplexity changed. Recomputing affinity matrix...", self.verbose
         ):
             self.P = joint_probabilities_nn(
-                self.__neighbors[:, :k_neighbors],
-                self.__distances[:, :k_neighbors],
```
pavlin-policar, line 192: I think this might be the reason for one of the failing tests. If we change the perplexity from 100 to 30, for instance, then `self.__neighbors` and `self.__distances` will still be N x 300, but we only want to calculate the P matrix for the 90 neighbors.

openTSNE/affinity.py
dkobak: Okay - but you don't allow this in `set_perplexity()`. You want to allow increasing the size of the kNN graph in `to_new()` but not in `set_perplexity()`?

openTSNE/affinity.py
```diff
@@ -275,7 +295,7 @@ def build_knn_index(
     data, method, k, metric, metric_params=None, n_jobs=1, random_state=None, verbose=False
 ):
     preferred_approx_method = nearest_neighbors.Annoy
-    if is_package_installed("pynndescent") and sp.issparse(data) and metric not in [
+    if is_package_installed("pynndescent") and (sp.issparse(data) or metric not in [
```
pavlin-policar, line 298: This snuck in here somehow.

pavlin-policar (commented):
I think the general idea should be fine after addressing these comments.

openTSNE/affinity.py
pavlin-policar: Yes, that's right. It doesn't make that much sense, I know, but I added this when playing around with your perplexity annealing idea. I didn't see a reason to allow increasing the perplexity at the time (and I still don't, really). I don't think this is a function that gets used very much anyways.

openTSNE/affinity.py
dkobak: I added it on purpose because I thought otherwise there will be a conflict (as I am modifying the same file in this PR). Not the proper way to handle it?

openTSNE/affinity.py
dkobak: On reflection, I see the reason: in `set_perplexity()` we don't need to run `query()` but in `to_new` we do. I'll fix.

openTSNE/affinity.py
dkobak: I am using this variable in `check_perplexity()`...

pavlin-policar: Oh, that's right! In this case, we could potentially allow increasing the perplexity! For instance, if we construct the Affinity model with perplexity=30 and k=300, calling `set_perplexity(100)` should work fine. So there still shouldn't be any dependence on `self._k_neighbors`. The dependence should be on `self.__neighbors.shape[1]`.

The key thing with `set_perplexity` was to not recompute the distance matrix. Everything that we can do with the existing distance matrix should be allowed.

Come to think of it, calling `set_perplexity(300)` should technically also work fine, since we're allowing that in the constructor. This is getting confusing. However, if we instantiate `a = Aff(perplexity=100, k_neighbors=300)`, and run that for a bit, then call `a.set_perplexity(250)` and run that for a bit. The embedding won't change much, since we're still working with the same KNNG. Which will be even more confusing.

However, I think consistency is more important, so if we allow `perplexity=250, k=300` in the constructor, this should definitely also be allowed using `set_perplexity`.

openTSNE/affinity.py
pavlin-policar: The easiest way would probably be to remove the change here and rebase the branch. And force push here. As it is now, we'll have merge conflicts. As long as you remove this change before you rebase, I think you'll have a clean rebase without any issues.

openTSNE/affinity.py
dkobak: I think my current implementation allows that. But adding a warning is a good idea. 

To be honest, I am not a fan of `check_perplexity()` setting perplexity automatically if the provided value is too high. Why not simply exit with error?

pavlin-policar: Oh, I missed the change in `check_perplexity`. That's not okay then, because `to_new` also calls `check_perplexity`, and it would be dependent on the original `k_neighbors`, which we don't want.

The purpose of `check_perplexity` is to make sure that the perplexity is within a valid range. With this change, we're allowing perplexity to be anything from 1 to K. And K should be anything between 1 to N. So maybe it would make more sense to change the function to accept both k and perplexity as parameters, and check if they're within these bounds. Then, in standard t-SNE, we'd call them using `check_perplexity(perplexity, N)`, and in transform `check_perplexity(transform_perplexity, N_reference_samples)`.

openTSNE/affinity.py
dkobak: OK removed from here.

openTSNE/affinity.py
dkobak: Fixed this.

openTSNE/affinity.py
dkobak: OK sounds good. I tried to implement this.

openTSNE/affinity.py
dkobak: Fixed.

pavlin-policar:
I think the implementation, as it stands, is pretty much fine now.

However, the same functionality should then be added to `to_new`. So, `to_new` should also have a parameter `k_neighbors` that might be "auto", or a numeric value. I also noticed you called `check_perplexity(..., np.inf)` here. It would make more sense to remember how many samples were in the original data and limit `k` to that. So for instance, if the embedding contains 300 data points, then anything above `k=300` should fail.

Regarding the tests -- I think they have to be fixed. From what I can tell, they are failing because before we weren't allowing `perplexity > k/3`. For instance
```
        aff = PerplexityBasedNN(self.x, perplexity=40)
       self.assertEqual(aff.perplexity, 30) # AssertionError: 40 != 30
```
this test expected the code to automatically reduce the perplexity to N/3 because N=91 here. But in this new implementation, that's completely fine, so the test needs to be changed. Same with the other test. It would be nice if you added some tests to check this new functionality, maybe with some of the examples I mentioned in this PR, but if not that's fine too. I can add them later on.

dkobak:
> However, the same functionality should then be added to to_new.

Fair enough - done.

dkobak:
I fixed those two tests, but now suddenly everything is failing -- I am not quite sure why.

PR #159, opened 2021-03-06
Approx bug fix

### LINKED ISSUE #158: Sparse input with cosine metric does not use Pynndescent
Sparse input should default to Pynndescent. But it doesn't if the metric is cosine. I don't understand why?

https://github.com/pavlin-policar/openTSNE/blob/master/openTSNE/affinity.py#L278

```
preferred_approx_method = nearest_neighbors.Annoy
    if is_package_installed("pynndescent") and sp.issparse(data) and metric not in [
        "cosine",
        "euclidean",
        "manhattan",
        "hamming",
        "dot",
        "l1",
        "l2",
        "taxicab",
    ]:
        preferred_approx_method = nearest_neighbors.NNDescent
```

It's weird but I am actually pretty sure that it used to work correctly. My code that used to work previously does not work anymore. Anyway, what's the reason for this `metric not in` clause?

CC @TimRepke.

dkobak:
Fixes #158.

### DIFF
--- a/openTSNE/affinity.py
+++ b/openTSNE/affinity.py
@@ -115,6 +115,10 @@ class PerplexityBasedNN(Affinities):
         number generator is the RandomState instance used by `np.random`.
 
     verbose: bool
+    
+    k_neighbors: int or ``auto``
+        The number of neighbors to use in the kNN graph. If ``auto`` (default),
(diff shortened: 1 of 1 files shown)


dkobak:
Oh boy, I messed up my branches.

Issue #162, opened 2021-03-09
Why does openTSNE use method=ball_tree and not method=auto for exact kNN?


dkobak:
I realized that sklearn uses brute-force kNN *always* whenever dimensionality `d>15`, see https://scikit-learn.org/stable/modules/neighbors.html.

Why not have `method='auto'` in https://github.com/pavlin-policar/openTSNE/blob/master/openTSNE/nearest_neighbors.py#L107 and rely on sklearn to choose the method?

dkobak:
Hmm, tried it out briefly and am not sure that their heuristic makes sense :-(

```
from sklearn.neighbors import NearestNeighbors

neigh = NearestNeighbors(n_neighbors=90, algorithm='ball_tree', n_jobs=-1)
%time neigh.fit(X[:10000, :784])
%time neigh.kneighbors_graph(mode='distance')

neigh = NearestNeighbors(n_neighbors=90, algorithm='auto', n_jobs=-1)
%time neigh.fit(X[:10000, :784])
%time neigh.kneighbors_graph(mode='distance')
```

Times:
```
Wall time: 695 ms
Wall time: 21 s
Wall time: 774 ms
Wall time: 47.2 s
```

The same for 20k:

```
Wall time: 1.65 s
Wall time: 1min 38s
Wall time: 2.24 s
Wall time: 2min 58s
```

Now I am very confused. They changed it relatively recently based on the benchmarks that showed that brute-force works faster...

dkobak:
I am going to close this issue... But if you have any thoughts, let me know please!

dkobak:
Update: turns out my sklearn was 0.23 and not 0.24, and after installing the latest version brute-force became MUCH MUCH faster! https://github.com/scikit-learn/scikit-learn/issues/10044#issuecomment-794407653

Here are updated times for full MNIST:

```
CPU times: user 23.3 s, sys: 83.6 ms, total: 23.4 s
Wall time: 23.3 s
CPU times: user 3h 9min 11s, sys: 11 s, total: 3h 9min 22s
Wall time: 25min 18s
CPU times: user 26.2 ms, sys: 0 ns, total: 26.2 ms
Wall time: 25.9 ms
CPU times: user 14min 21s, sys: 1min 40s, total: 16min 1s
Wall time: 3min 48s
```

I guess we should change it in openTSNE. Re-opening.

pavlin-policar:
[...]

PR #163, opened 2021-03-10
Add sklearn nearest_neighbor search with method="auto"

pavlin-policar:
##### Issue
#162

##### Description of changes
I initially thought we could get rid of all the cosine distance logic, but it turns out that scikit-learn doesn't implement it. So that workaround stays. Basically, this is just a copy of `BallTree`, but with `method="auto"`.

I've also deprecated `BallTree`, which will be removed in future versions. Then, we can also get rid of this code duplication.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/openTSNE/affinity.py
+++ b/openTSNE/affinity.py
@@ -288,12 +288,12 @@ def build_knn_index(
         preferred_approx_method = nearest_neighbors.NNDescent
 
     if data.shape[0] < 1000:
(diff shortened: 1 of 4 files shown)


dkobak:
Should the `neighbors="exact"` now call the new class?

pavlin-policar:
Absolutely. I missed that. Thanks!

PR #175, opened 2021-04-16
Refactor affinity module

pavlin-policar:
##### Description of changes
Adding this kind of functionality complicated things quite a bit, so a pretty major refactor was necessary to keep things manageable. Notable changes include:
- `build_knn_index` no longer computes the neighbors, but only returns the correct index. This lets us pass in already existing `KNNIndex` objects to affinities.
- All `Affinity` classes can now be called in two ways. 1) with a data matrix, as is standard, or 2) using an existing `KNNIndex` object. This is in preparation for scanpy integration, where we will have to use existing KNNGs.
- `KNNIndex` objects now have the default parameter `metric="euclidean". This just makes it easier to debug or instantiate manually.
- Add `nearest_neighbors.PrecomputedNeighbors` which allows us to wrap the standard KNNG representations.

Because there are now a whole bunch of possible ways to call this API, I added a test module which runs over the different ways to create an iris embedding. See `tests/test_different_usages.py`.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/openTSNE/affinity.py
+++ b/openTSNE/affinity.py
@@ -1,6 +1,5 @@
 import logging
 import operator
-from typing import Iterable
 from functools import reduce
 
 import numpy as np
@@ -35,6 +34,7 @@ class Affinities:
     def __init__(self, verbose=False):
         self.P = None
         self.verbose = verbose
+        self.knn_index: nearest_neighbors.KNNIndex = None
 
     def to_new(self, data, return_distances=False):
         """Compute the affinities of new samples to the initial samples.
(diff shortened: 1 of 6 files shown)


dkobak:
Wow, this `precomputed` thing escalated pretty quickly :)

pavlin-policar:
Haha, yeah. Things got pretty complicated when I just patched things up. So I tried to simplify it, but all the different ways we can call things now and string them together have made things a bit messy and necessarily complicated.

I'm trusting that my tests are thorough enough that nothing broke here. Most of the major things are tested pretty well though.

Issue #216, opened 2022-10-07
Create an affinity object with precomputed affinity matrix


dkobak:
If I have a sparse affinity matrix stored in `P`, I can run openTSNE as follows:

```
from openTSNE import TSNE
from openTSNE.affinity import Affinities

A = Affinities()
A.P = P
Z = TSNE().fit(affinities = A)
```

I was wondering it would make sense to implement the following API:
```
Z = TSNE().fit(affinities = Affinities(P))
```

pavlin-policar:
Hmm, the `Affinities` object is really meant more of as a base-class, and I never intended it to be used in this way. But I suppose there is no reason this should not work fine. There are two options, I think. One is the API you suggest `Affinities(P)`, but I think making a new class doing basically the same thing called `PrecomputedAffinities(P)` would fit more nicely into the API. Since we already have the `PrecomputedNeighbors` KNN index. That would definitely make sense. 

Since you already have a use-case in mind, would you be willing to take a crack at it?

PR #217, opened 2022-10-12
Implement precomputed affinity class

dkobak:
Implements #216.

Allows to do the following:
```python
from openTSNE import TSNE
from openTSNE.affinity import PrecomputedAffinities

# Here P is either a np.array or a scipy.sparse matrix
Z = TSNE().fit(affinities = PrecomputedAffinities(P))
```

### DIFF
--- a/openTSNE/affinity.py
+++ b/openTSNE/affinity.py
@@ -1264,3 +1264,22 @@ class Uniform(Affinities):
             return P, neighbors, distances
 
         return P
+
+
+class PrecomputedAffinities(Affinities):
+    """Use a precomputed affinity matrix.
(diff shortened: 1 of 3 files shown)


pavlin-policar:
I think this functionality needs a test. Additionally, would it make sense to re-normalize the affinity matrix here? Or do we want to allow un-normalized affinity matrices? Perhaps a `normalize` parameter defaulting to `True` might be the way to go. What do you think?

pavlin-policar:
Wonderful, thanks! I've added a usage-test, which basically just checks that the whole thing works as intended, then I'll merge.

Issue #218, opened 2022-10-12
Better support for large late exaggeration


dkobak:
I am often using `exaggeration` argument of `TSNE()`. It works flawlessly when the argument is less than 12:

```
TSNE(exaggeration=4).fit(X)
```

But if I want to use a value above 12, e.g. 24 (this rarely happens but I ran into this situation just now), then I am doing smth like this:

```
TSNE(exaggeration=24, early_exaggeration=24, learning_rate = X.shape[0]/24).fit(X)
```

which is a little annoying, and cumbersome to communicate to practitioners who are less familiar with inner working of t-SNE. I would love to be able to simply run

```
TSNE(exaggeration=24).fit(X)
```

I am wondering if we can somehow change the default behaviour, so that this works by default. For example, we could have `early_exaggeration='auto'` by default, so that `auto` defaults to 12 unless exaggeration is above 12. And set the "auto" learning rate accordingly. This is admittedly an edge use case... But I don't really see downsides. What do you think?

pavlin-policar:
I see, yes, this makes sense. I think the "auto" route is the way to go. It is an unusual use-case, but I don't really see any downside to this approach.

PR #220, opened 2022-11-09
Implement auto early exaggeration

dkobak:
Implements #218.

First, `early_exaggeration="auto"` is now set to `max(12, exaggeration)`.

Second, the learning rate. We have various functions that currently take `learning_rate="auto"` and set it to `max(200, N/12)`. I did not change this, because those functions usually do not know what the early exaggeration was. So I kept it as is. I only changed the behaviour of the base class: there `learning_rate="auto"` is now set to `max(200, N/early_exaggeration)`.

This works as intended:

```Python
X = np.random.randn(10000,10)

TSNE(verbose=True).fit(X)
# Prints 
# TSNE(early_exaggeration=12, verbose=True)
# Uses lr=833.33

TSNE(verbose=True, exaggeration=5).fit(X)
# Prints
# TSNE(early_exaggeration=12, exaggeration=5, verbose=True)
# Uses lr=833.33

TSNE(verbose=True, exaggeration=20).fit(X)
# Prints
# TSNE(early_exaggeration=20, exaggeration=20, verbose=True)
# Uses lr=500.00
```

(Note that the learning rate is currently not printed by the `repr(self)` because it's kept as "auto" at construction time and only set later. That's also how we had it before.)

### DIFF
--- a/openTSNE/tsne.py
+++ b/openTSNE/tsne.py
@@ -1007,7 +1007,7 @@ class TSNE(BaseEstimator):
         The exaggeration factor to use during the *early exaggeration* phase.
         Typical values range from 4 to 32. When ``early_exaggeration="auto"``
         early exaggeration factor defaults to 12, unless desired subsequent
-        exaggeration is higher, i.e.: ``early_exaggeration = min(12,
+        exaggeration is higher, i.e.: ``early_exaggeration = max(12,
         exaggeration)``.
 
     n_iter: int
@@ -1152,7 +1152,10 @@ class TSNE(BaseEstimator):
         self.perplexity = perplexity
         self.learning_rate = learning_rate
         if early_exaggeration == "auto":
-            self.early_exaggeration = min(12, exaggeration)
+            if exaggeration is None:
+                self.early_exaggeration = 12
+            else:
+                self.early_exaggeration = max(12, exaggeration)
         else:
             self.early_exaggeration = early_exaggeration
         self.early_exaggeration_iter = early_exaggeration_iter
--- a/openTSNE/tsne.py
+++ b/openTSNE/tsne.py
@@ -88,8 +88,13 @@ def _handle_nice_params(embedding: np.ndarray, optim_params: dict) -> None:
     optim_params["n_jobs"] = n_jobs
 
     # Determine learning rate if requested
-    if optim_params.get("learning_rate", "auto") == "auto":
-        optim_params["learning_rate"] = max(200, n_samples / 12)
+    learning_rate = optim_params.get("learning_rate", "auto")
+    if learning_rate == "auto":
+        exaggeration = optim_params.get("exaggeration", None)
+        if exaggeration is None:
+            exaggeration = 1
+        learning_rate = n_samples / exaggeration
+    optim_params["learning_rate"] = learning_rate
 
 
 def __check_init_num_samples(num_samples, required_num_samples):
@@ -160,11 +165,12 @@ class PartialTSNEEmbedding(np.ndarray):
 
     learning_rate: Union[str, float]
         The learning rate for t-SNE optimization. When ``learning_rate="auto"``
-        the appropriate learning rate is selected according to max(200, N / 12)
+        the appropriate learning rate is selected according to N / exaggeration
         as determined in Belkina et al. (2019), Nature Communications. Note that
-        this should *not* be used when adding samples into existing embeddings,
-        where the learning rate often needs to be much lower to obtain
-        convergence.
+        this will result in a different learning rate during the early
+        exaggeration phase and afterwards. This should *not* be used when 
+        adding samples into existing embeddings, where the learning rate often
+        needs to be much lower to obtain convergence.
 
     exaggeration: float
         The exaggeration factor is used to increase the attractive forces of
@@ -278,14 +284,14 @@ class PartialTSNEEmbedding(np.ndarray):
             The number of optimization iterations.
 
         learning_rate: Union[str, float]
-            The learning rate for t-SNE optimization. When
+            The learning rate for t-SNE optimization. When 
             ``learning_rate="auto"`` the appropriate learning rate is selected
-            according to max(200, N / 12), as determined in Belkina et al.
-            "Automated optimized parameters for t-distributed stochastic
-            neighbor embedding improve visualization and analysis of large
-            datasets", 2019. Note that this should *not* be used when adding
-            samples into existing embeddings, where the learning rate often
-            needs to be much lower to obtain convergence.
+            according to N / exaggeration as determined in Belkina et al.
+            (2019), Nature Communications. Note that this will result in a
+            different learning rate during the early exaggeration phase and
+            afterwards. This should *not* be used when adding samples into
+            existing embeddings, where the learning rate often needs to be much
+            lower to obtain convergence.
 
         exaggeration: float
             The exaggeration factor is used to increase the attractive forces of
@@ -414,10 +420,12 @@ class TSNEEmbedding(np.ndarray):
 
     learning_rate: Union[str, float]
         The learning rate for t-SNE optimization. When ``learning_rate="auto"``
-        the appropriate learning rate is selected according to max(200, N / 12),
-        as determined in Belkina et al. "Automated optimized parameters for
-        T-distributed stochastic neighbor embedding improve visualization and
-        analysis of large datasets", 2019.
+        the appropriate learning rate is selected according to N / exaggeration
+        as determined in Belkina et al. (2019), Nature Communications. Note that
+        this will result in a different learning rate during the early
+        exaggeration phase and afterwards. This should *not* be used when 
+        adding samples into existing embeddings, where the learning rate often
+        needs to be much lower to obtain convergence.
 
     exaggeration: float
         The exaggeration factor is used to increase the attractive forces of
@@ -552,12 +560,14 @@ class TSNEEmbedding(np.ndarray):
             The number of optimization iterations.
 
         learning_rate: Union[str, float]
-            The learning rate for t-SNE optimization. When
+            The learning rate for t-SNE optimization. When 
             ``learning_rate="auto"`` the appropriate learning rate is selected
-            according to max(200, N / 12), as determined in Belkina et al.
-            "Automated optimized parameters for t-distributed stochastic
-            neighbor embedding improve visualization and analysis of large
-            datasets", 2019.
+            according to N / exaggeration as determined in Belkina et al.
+            (2019), Nature Communications. Note that this will result in a
+            different learning rate during the early exaggeration phase and
+            afterwards. This should *not* be used when adding samples into
+            existing embeddings, where the learning rate often needs to be much
+            lower to obtain convergence.
 
         exaggeration: float
             The exaggeration factor is used to increase the attractive forces of
@@ -693,7 +703,7 @@ class TSNEEmbedding(np.ndarray):
         early_exaggeration_iter=0,
         exaggeration=1.5,
         n_iter=250,
-        initial_momentum=0.5,
+        initial_momentum=0.8,
         final_momentum=0.8,
         max_grad_norm=0.25,
         max_step_norm=None,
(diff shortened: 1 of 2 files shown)


openTSNE/tsne.py
```diff
@@ -1395,11 +1404,17 @@ def prepare_initial(self, X=None, affinities=None, initialization=None):
...
+        if self.learning_rate == "auto":
+            learning_rate_now = max(200, n_samples / self.early_exaggeration)
+        else:
+            learning_rate_now = self.learning_rate 
 
         gradient_descent_params = {
             "dof": self.dof,
             "negative_gradient_method": self.negative_gradient_method,
-            "learning_rate": self.learning_rate,
+            "learning_rate": learning_rate_now,
```
pavlin-policar, line 1417: I'm not sure this is the correct put to put this. This would set the learning rate for the duration of the whole embedding, not just the early exaggeration phase, right? I think the best place to put this might actually be inside the `gradient_descent` call function, so the correct rescaling will happen during any call to `.optimize`, not just the standard `TSNE().fit` call.

I wasn't aware that this was an issue during the early exaggeration phase at all. It is actually necessary to rescale the learning rate, if the exaggeration isn't 1? In that case, wouldn't it also make sense to rescale it given _any_ exaggeration value, not just the early exaggeration phase?

dkobak: The current default openTSNE behaviour (which is the same in FIt-SNE and about to become default in soon-to-be-released sklearn 1.2) is to use learning rate N/12. Here 12 corresponds to the early exaggeration, the idea being that if learning_rate * exaggeration > 1, then the optimization may diverge (as shown by Linderman & Steinerberger). One COULD use learning rate N, instead of N/12, for the 2nd optimization phase, once early exaggeration is turned off. But we do not do it. We could decide to do it, and I think it may even be a good idea, but this will make openTSNE default behaviour different from other implementations. In any case, this I think requires a different PR and a separate issue, maybe comparing current and suggested scheme on a range of datasets.

In this PR I assumed that we want to keep the learning rate constant throughout optimization. Which is why I set it to N/early_exaggeration and keep the same learning rate for the second phase.

pavlin-policar: Yes, that makes sense, I'd forgotten about that. However, I'm a bit unsure what the best way about implementing would be. We currently handle `learning_rate="auto"` [here](https://github.com/pavlin-policar/openTSNE/blob/master/openTSNE/tsne.py#L90-L92) in the `_handle_nice_params` function. However, this function is called in every `.optimize` step, so that we can perform whatever optimization sequence we want.

With the way you've implemented it now, we'd be going around this function because we'd change the `learning_rate` from `auto` to a number before we even got to this function. This would also cause inconsistent behaviour between
```python
TSNE().fit()
```
and
```python
embedding.optimize(...)
embedding.optimize(...)
```
since the first one would then use the "correct" learning rate, while the second would use `N/12` in both cases. I am very much opposed to this inconsistency, but I'm not sure what the best way to handle this would be.

Perhaps the best course of action would be to investigate rescaling the learning rate based on exaggeration, so we'd end up with a different learning rate for the early exaggeration and the standard optimization phase. I think this would best fit into our framework, and it seems the most principled approach.

dkobak: I see your point. I will start a separate issue.

pavlin-policar:
To build on top of https://github.com/pavlin-policar/openTSNE/pull/220#discussion_r1017919909, I've run a quick experiment to see convergence rates using lr/exaggeration, and the results indicate that N/exag may actually lead to better and faster convergence than simply dividing by N/ee_rate:

image

Visually, all four embeddings look pretty similar, and I wouldn't say ones are less converged than the others, but it seems like we can get the same KL divergence faster this way. What do you think?

dkobak:
Does `exag` in this plot mean early exaggeration? And late exaggeration is set to 1? Is this MNIST?

pavlin-policar:
> Does exag in this plot mean early exaggeration?
Yes, the "regular" phase is run with exaggeration=1

> Is this MNIST?
No, this is my typical macosko example.

This is one example, so we'd definitely have to check it on several more, but this does indicate "that at least it wouldn't hurt". And if we were to implement N/exag, this would fit in much more cleanly into the entire openTSNE architecture. I wouldn't mind being slightly inconsistent with FIt-SNE or scikit-learn in this respect, since the visualizations seem visually indistinguishable from one another.

dkobak:
I agree. I ran the same test on MNIST, and observe faster convergence using the suggested approach.

Incidentally, I first did it wrong because I did not specify correct momentum terms for the two optimize() calls. This made me think that I am not aware of any reason for why momentum should be different. I tried using momentum=0.8 for both stages, and it seems to be better than the current 0.5->0.8 scheme. 

Note that your criticism that `TSNE().fit()` and twice calling `embedding.optimize(...)` is not identical, also applies to momentum differences, no?

optimization-learing-rate

pavlin-policar:
> Note that your criticism that TSNE().fit() and twice calling embedding.optimize(...) is not identical, also applies to momentum differences, no?

Yes, this is the same issue, and this has bothered me from the very start. So I'm very happy to see that using a single momentum seems to lead to faster convergence, as this would justify defaulting to 0.8 everywhere.

I'm not aware of any justification for this choice either, this came from the original LVM and I never bothered to question it.

dkobak:
It seems that 250 iterations may be too many with this momentum setting, but maybe let's not touch the n_iter defaults for now.

Would be good to check this on a couple of more datasets, perhaps also very small ones (Iris?), but overall I think it looks good.

pavlin-policar:
> It seems that 250 iterations may be too many with this momentum setting, but maybe let's not touch the n_iter defaults for now.

Yes, I agree.

> Would be good to check this on a couple of more datasets, perhaps also very small ones (Iris?), but overall I think it looks good.

I'd also check a couple of big ones, the cao one, maybe the 10x mouse as well. It might also be interesting to see if we actually need lr=200 on iris. Maybe lr=N=150 would be better. The 200 now seems kind of arbitrary.

dkobak:
Iris:

optimization-learing-rate-iris

> I'd also check a couple of big ones, the cao one, maybe the 10x mouse as well. It might also be interesting to see if we actually need lr=200 on iris. Maybe lr=N=150 would be better. The 200 now seems kind of arbitrary.

Very good point. Red line shows that turning off learning rate "clipping" (I mean clipping to 200) works actually very good.

dkobak:
Are you going to run it on smth with sample size over 1mln? Sounds like you have everything set up for these experiments. But if you want, I can run something as well.

pavlin-policar:
Yes, sure, I'll find a few more data sets and run them. If everything goes well, we'll change the defaults to momentum=0.8 and lr=N/exaggeration, and this will solve all the issues outlined above.

dkobak:
This may be worth bumping to 0.7!

dkobak:
Have you had a chance to run it on some other datasets? Otherwise I would give it a try on something large, I am curious :)

pavlin-policar:
Hey Dmitry, no, unfortunately, I haven't had time yet. It's a busy semester for teaching :) If you have any benchmarks you'd like to run, I'd be happy to see the results.

dkobak:
Just ran it in an identical way for Iris, MNIST, and n=1.3 mln dataset from 10x. I used uniform affinities with k=10 in all cases, to speed things up.

Old: current default
New: learning rate N/12, followed by learning rate N, momentum always .8, learning rate below 200 is allowed

optimization-learing-rate

I think everything is very consistent:

* Learning rate N makes convergence in the 2ns phase faster
* Momentum .8 makes convergence in the 1st phase faster
* Learning rate below 200 for small datasets prevents fluctuations in loss

pavlin-policar:
I've also tried this on two other data sets: shekhar (20k) and cao (2mln):

shekhar (20k):
image

cao (2mln):
image

In both cases, using `momentum=0.8` and `learning_rate=N/exaggeration` works better. Along with the the other examples you provide, I feel this is sufficient to change the defaults.

> Learning rate below 200 for small datasets prevents fluctuations in loss

I don't understand this entirely. So, can we use `learning_rate=N/exaggeration` in all cases? Or do we keep it as it is right now and use `learning_rate=max(N/exaggeration, 200)`?

dkobak:
> I don't understand this entirely. So, can we use learning_rate=N/exaggeration in all cases? Or do we keep it as it is right now and use learning_rate=max(N/exaggeration, 200)?

I now think that we *should* use `learning_rate=N/exaggeration in all cases`. For iris (N=150), this would mean learning rate 150/12 = 12.5 during the early exaggeration phase. Currently we use learning rate 200 during that phase, which is too high. It does not lead to divergence (not sure why; maybe due to some gradient or step size clipping?) but does lead to oscillating loss, clearly suggesting that something is not well with the gradient descent. Learning rate 12.5 seems much more stable.

pavlin-policar:
Yes, I agree. I tried it myself and there are big oscillations. E.g. subsampling iris to 50 data points also leads to less oscillation with `learning_rate=N/exaggeration`.

I think the best course of action is to add a `learning_rate="auto"` option, make that the default, and then handle it in `_handle_nice_params`, as I've written in the original code review.

dkobak:
Do you want me to edit this PR, or do you prefer to make a new one?

> I think the best course of action is to add a learning_rate="auto" option, make that the default, and then handle it in _handle_nice_params, as I've written in the original code review.

Is it okay to make `_handle_nice_params` take the exaggeration value as input? Currently it does not.

pavlin-policar:
I think adding it to this PR is completely fine.

> Is it okay to make _handle_nice_params take the exaggeration value as input? Currently it does not.

I think it does. The exaggeration factor should come in through [`.optimize`](https://github.com/pavlin-policar/openTSNE/blob/46d65aec8e299f1152511004e8efa34c823510af/openTSNE/tsne.py#L266), and should be captured in the `**gradient_descent_params`. Then, this should be passed through [here](https://github.com/pavlin-policar/openTSNE/blob/46d65aec8e299f1152511004e8efa34c823510af/openTSNE/tsne.py#L374-L377).

dkobak:
But exaggeration factor does not really feel like a "gradient descent parameter"... So I was reluctant to add it into the `gradient_descent_params`. What about passing it into `_handle_nice_params()` as an additional separate input parameter? Like this:

```
def _handle_nice_params(embedding: np.ndarray, exaggeration: double, optim_params: dict) -> None:
```

pavlin-policar:
My understanding is that it is already included in the `_handle_nice_params`. Indeed, if I run a simple example and print the `optim_params` in `_handle_nice_params`, I get

```
{'learning_rate': 'auto', 'momentum': 0.5, 'theta': 0.5, 'max_grad_norm': None, 'max_step_norm': 5,
'n_jobs': 1, 'verbose': False, 'callbacks': None, 'callbacks_every_iters': 50,
'negative_gradient_method': 'bh', 'n_interpolation_points': 3, 'min_num_intervals': 50,
'ints_in_interval': 1, 'dof': 1, 'exaggeration': 12, 'n_iter': 25}
```
for the EE phase, and
```
{'learning_rate': 'auto', 'momentum': 0.8, 'theta': 0.5, 'max_grad_norm': None, 'max_step_norm': 5,
'n_jobs': 1, 'verbose': False, 'callbacks': None, 'callbacks_every_iters': 50,
'negative_gradient_method': 'bh', 'n_interpolation_points': 3, 'min_num_intervals': 50,
'ints_in_interval': 1, 'dof': 1, 'exaggeration': None, 'n_iter': 50}
```
for the standard phase. Importantly, `exaggeration` and the `learning_rate` is already among them. So I would imagine something like this would be totally fine:
```python
learning_rate = optim_params["learning_rate"]
if learning_rate == "auto":
    exaggeration = optim_params.get("exaggeration", None)
    if exaggeration is None:
        exaggeration = 1
    learning_rate = n_samples / exaggeration
optim_params["learning_rate"] = learning_rate
```

dkobak:
I see. Still not quite sure where it gets added to the dictionary, but it does not matter now.

I did the changes and added a test that should check if running `optimize()` twice produces the same result as running `transform()`, however this test fails, and I am not yet sure why :-/ Here is what I want to be the same:

```
from sklearn import datasets
iris = datasets.load_iris()
X = iris.data

embedding1 = TSNE(verbose=True).fit(X)
        
A = affinity.PerplexityBasedNN(X, verbose=True)
I = initialization.pca(X, verbose=True)
embedding2 = TSNEEmbedding(I, A, verbose=True)
embedding2.optimize(n_iter=250, exaggeration=12, verbose=True)
embedding2.optimize(n_iter=500, verbose=True)
```

Here is the output that I get:

```
===> Running optimization with exaggeration=12.00, lr=12.50 for 250 iterations...
Iteration   50, KL divergence 0.9020, 50 iterations in 0.0314 sec
Iteration  100, KL divergence 0.9095, 50 iterations in 0.0255 sec
Iteration  150, KL divergence 0.9027, 50 iterations in 0.0250 sec
Iteration  200, KL divergence 0.8984, 50 iterations in 0.0251 sec
Iteration  250, KL divergence 0.8955, 50 iterations in 0.0247 sec
===> Running optimization with exaggeration=1.00, lr=150.00 for 500 iterations...
Iteration   50, KL divergence 0.1413, 50 iterations in 0.0247 sec
Iteration  100, KL divergence 0.1292, 50 iterations in 0.0243 sec
...

===> Running optimization with exaggeration=12.00, lr=12.50 for 250 iterations...
Iteration   50, KL divergence 0.9020, 50 iterations in 0.0249 sec
Iteration  100, KL divergence 0.9095, 50 iterations in 0.0263 sec
Iteration  150, KL divergence 0.9027, 50 iterations in 0.0255 sec
Iteration  200, KL divergence 0.8984, 50 iterations in 0.0259 sec
Iteration  250, KL divergence 0.8955, 50 iterations in 0.0253 sec
===> Running optimization with exaggeration=1.00, lr=150.00 for 500 iterations...
Iteration   50, KL divergence 0.1383, 50 iterations in 0.0245 sec
Iteration  100, KL divergence 0.1304, 50 iterations in 0.0240 sec
```

So the early exaggeration phase seems to be identical, but afterwards the results diverge...

dkobak:
Okay, never mind, I figured it out... Fixing.

pavlin-policar:
I think this is fine now. I'm glad we found a way to simplify the API and speed up convergence at the same time :)

Issue #223, opened 2022-12-16
Switching spectral initialization to sklean.manifold.SpectralEmbeddings


dkobak:
A student in our lab is currently looking into spectral initialization, and she found out that `openTSNE.init.spectral(tol=0)` in some cases does not agree to `sklearn.manifold.SpectralEmbedding(affinity='precomputed')`. In some cases it does agree perfectly or near-perfectly, but we have an example when the result is very different, and `SpectralEmbedding` gives what seems like a more meaningful result.

I looked at the math, and it seems that they should conceptually be computing the same thing (`SpectralEmbedding` finds eigenvectors of the L_sym, whereas `init.spectral` finds generalized eigenvectors or W and D, but that should be equivalent, as per https://jlmelville.github.io/smallvis/spectral.html @jlmelville). We don't know what the difference is due to. It may be numerical. 

However, conceptually, it seems sensible if openTSNE would simply outsource the computation to sklearn.

A related issue is that `init.spectral` is not reproducible and gives different results with each run. Apparently the way we initialize `v0` makes ARPACK to still have some randomness. Sklearn gets around this by initializing `v0` differently. I guess openTSNE should do the same -- but of course if we end up simply calling `SpectralEmbedding` then it won't matter.

jlmelville:
@dkobak don't want to hijack this issue so feel free to contact me directly, but I am curious:

* is the example a simulation dataset or have any other unusually regular structure?
* do you know what codepath and solver the [spectral embedding](https://github.com/scikit-learn/scikit-learn/blob/dc580a8ef5ee2a8aea80498388690e2213118efd/sklearn/manifold/_spectral_embedding.py) ends up using?
* have you been able to look at the eigenvalues that come out? No chance that the solver has accidentally skipped an eigenvalue which just happens to give a nicer looking embedding? The discussion at https://github.com/yixuan/spectra/issues/126 suggests that for theoretical reasons with shift-and-invert (which spectral embedding uses with the codepath that ends up calling `eigsh`), if the shift value matches the eigenvalue then it will be skipped. I can't say that arpack's version suffers from that but might be worth checking.
* any noticeable differences in run time?
[...]

Issue #180, opened 2021-04-18
Memory collapses with precomputed block matrix


fsvbach:
##### Expected behaviour

When I run tSNE on a symmetric 200x200 block matrix such as this one
![distancematrix0](https://user-images.githubusercontent.com/64090304/115153699-9ff33280-a077-11eb-8e9b-fa457b0b9abc.png)
I expect TSNE to return 4 distinct clusters (actually 4 points only).  Sklearn yields this. 

##### Actual behaviour

Using openTSNE the terminal crashes with full memory (50% of the time). If it survives the clusters are visible, however the result is not as satisfying. 

##### Steps to reproduce the behavior

matrix = Block matrix
tsne = TSNE(metric='precomputed', initialization='spectral', negative_gradient_method='bh')
embedding = tsne.fit(matrix)

NOTE: I am using the direct installation from GitHub this morning.

pavlin-policar:
Please provide a minimal code example where this fails so we can reproduce and debug this. There really isn't anything obvious that would cause the memory to balloon.

Does the same thing happen when using `negative_gradient_method="fft"`?

fsvbach:
Okay, with negative_gradient_method="fft" it works, but I sometimes get the RuntimeWarning: invalid value encountered in log
  kl_divergence_ += sum_P * np.log(sum_Q + EPSILON)

A minimal example would be:
`
from openTSNE import TSNE
import numpy as np
matrix = np.genfromtxt('matrix.csv', delimiter=',')
tsne = TSNE(metric='precomputed', initialization='spectral', negative_gradient_method='bh')
embedding = tsne.fit(matrix)
`
with the matrix in matrix.csv 
[matrix.csv](https://github.com/pavlin-policar/openTSNE/files/6331881/matrix.csv)

Note: It doesn't crash always, but every third or fourth time, so I think its due to some initialization?

pavlin-policar:
What I suspect may be happening is that if the points are too close together in the embedding, the BH tree balloons. However, I should have some safeguards in place so that this doesn't happen. Maybe those safeguards aren't enough. This might be plausible since you're trying to force all the points super close together. But I'd need to investigate it further. I suggest you use `fft` for the time being.

dkobak:
> I expect TSNE to return 4 distinct clusters (actually 4 points only). Sklearn yields this.

[...]

PR #225, opened 2023-01-12
Improve PCA and spectral inits

### LINKED ISSUE #224: Adding tiny amount of noise to PCA/spectral init to prevent points from overlapping
A student encountered an unusual situation with spectral initialization: due to the specifics of the dataset, there were points that were EXACTLY overlapping in the initialization. This made the points "stuck" to each other forever -- these points felt the same repulsive force to all other points and so could not separate from each other, even though there should actually be repulsion between them. Adding a tiny amount of random noise to the initialization solved this problem and made points spread over the embedding, as expected.

This reminded me of another issue we discussed a while ago https://github.com/pavlin-policar/openTSNE/issues/180 (still open) where points exactly overlapping in the initialization were causing some problems.

My suggestion is to _always_ add a tiny amount of noise to all initializations that we compute. Specifically, in the `rescale()` function here https://github.com/pavlin-policar/openTSNE/blob/master/openTSNE/initialization.py#L9 I would replace

```
x /= np.std(x[:, 0]) * 10000
```

with

```
x /= np.std(x[:, 0]) * 10000
x += np.random.randn(*x.shape) * 1e-5
```

This would affect PCA and spectral init, but would *not* affect a user-provided init.

dkobak:
A tiny PR that would fix #224, #223, and #180. I am not insisting on any of these changes, but thought I make the PR to make it easier to discuss.

### DIFF
--- a/openTSNE/initialization.py
+++ b/openTSNE/initialization.py
@@ -6,13 +6,15 @@ from sklearn.utils import check_random_state
 from openTSNE import utils
 
 
-def rescale(x, inplace=False):
+def rescale(x, inplace=False, target_std=1e-4, add_noise=True, random_state=None):
     """Rescale an embedding so optimization will not have convergence issues.
 
     Parameters
     ----------
     x: np.ndarray
     inplace: bool
+    target_std: float
+    addNoise: bool
 
     Returns
     -------
@@ -23,7 +25,13 @@ def rescale(x, inplace=False):
     if not inplace:
         x = np.array(x, copy=True)
 
-    x /= np.std(x[:, 0]) * 10000
+    x /= np.std(x[:, 0]) / target_std
+    
+    # Add noise with standard deviation 100 times smaller
+    # to avoid numerical problems when the points overlap exactly
+    if add_noise:
+        random_state = check_random_state(random_state)
+        x += random_state.normal(0, target_std / 100, *x.shape)
 
     return x
 
@@ -143,7 +151,9 @@ def spectral(A, n_components=2, tol=1e-4, max_iter=None, random_state=None, verb
 
     # Find leading eigenvectors
     k = n_components + 1
-    v0 = np.ones(A.shape[0]) / np.sqrt(A.shape[0])
+    # v0 initializatoin is taken from sklearn.utils._arpack._init_arpack_v0()
+    random_state = check_random_state(random_state)
+    v0 = random_state.uniform(-1, 1, A.shape[0])
     eigvals, eigvecs = sp.linalg.eigsh(
         A, M=D, k=k, tol=tol, maxiter=max_iter, which="LM", v0=v0
     )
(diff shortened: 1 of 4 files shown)


dkobak:
A lot of tests fail because we would need to fix random seeds over there. I'm not doing it for now -- let's discuss first.

pavlin-policar:
I think adding a tiny amount of noise to each initialization is fine if it solves all these issues. I'm used to calling this *jitter*, so that's what I'll call it. However, I think the user should be able to turn this off. For instance, we could add a general parameter `TSNE(..., jitter_initialization=True`) so that the user can still turn this off. I would also put this into its own separate function e.g.

```python
def jitter(x, scale=0.01, random_state):
    std = np.std(x[:, 0])
    target_std = std * scale
    return x + random_state.normal(0, std, x.shape)
```

The we can call this `pca`, `spectral`, and any other initializations we have. I definitely wouldn't add this to the `rescale` function as that could be very confusing having `rescale` also actually change our initialization. We can turn this off by default for precomputed embeddings, but this might be a pain, because then we'd again need to implement something like `jitter_initialization="auto"` so we can figure out if we need to jitter or not. Maybe it would be easier to just enable jittering by default for all initializations, and if the user wants, they can turn this off manually. We'd need to document this behaviour in the initialization parameters.

dkobak:
Sure, let me move jitter out of `rescale()` into `jitter()`.

I also agree that there should be a way to switch it off. What about adding `add_jitter=True` parameter to `pca()` and to `spectral()`? Then the user would be able to switch jitter off by manually calling e.g. `pca(add_jitter=False)`.

I am not so sure about having the global parameter like `TSNE(..., jitter_initialization=True)`, because indeed I think it would make more sense not to add jitter to a user-provided initialization.

pavlin-policar:
> What about adding add_jitter=True parameter to pca() and to spectral()? Then the user would be able to switch jitter off by manually calling e.g. pca(add_jitter=False).

Yes, this is exactly what I had in mind. We should add this to both `pca` and `spectral`. The other two -- `weighted_mean` and `median` can only be used when adding new points to an embedding, and they ignore one another, so it doesn't make any difference if we jitter those, so I wouldn't add the jitter parameter to those.

> I am not so sure about having the global parameter like TSNE(..., jitter_initialization=True), because indeed I think it would make more sense not to add jitter to a user-provided initialization.

So you'd want to have jitter on by default for `pca` and `spectral`, but have it off for `precomputed`? The issue I see with that is that the user then can't use standard `pca` without jitter without having to precompute the initialization, then passing it through the constructor.

It might be fine though. The TSNE constructor has enough parameters as it is, and nobody will notice this in practice, since the added noise should be very small. Okay, I think this is fine for now then, and we'll add the parameter later, if we find any need for it.

dkobak:
Agree. I refactored the jitter code, and also added a test that checks that our spectral init coincides with sklearn spectral embedding. Luckily, it passes :-)

I also had to fix random seeds in some existing tests that used PCA/spectral initializations.

pavlin-policar:
Great, thanks! I'll close the related issues.

PR #226, opened 2023-01-23
azure-pipelines-release.yml: Build macos universal2 wheels

ales-erjavec:
##### Issue

Provide macos universal2 wheels for apple silicon

##### Description of changes

Build macos universal2 wheels.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/azure-pipelines-release.yml
+++ b/azure-pipelines-release.yml
@@ -21,28 +21,44 @@ jobs:
         osx - python37:
           image.name: 'macos-12'
           python.version: '3.7'
+          ciwb.build: cp37-macosx_x86_64
+          ciwb.archs: x86_64
         osx - python38:
           image.name: 'macos-12'
           python.version: '3.8'
+          ciwb.build: cp38-macosx_x86_64
+          ciwb.archs: x86_64
         osx - python39:
           image.name: 'macos-12'
           python.version: '3.9'
+          ciwb.build: cp39-macosx_universal2
+          ciwb.archs: universal2
         osx - python310:
           image.name: 'macos-12'
           python.version: '3.10'
+          ciwb.build: cp310-macosx_universal2
+          ciwb.archs: universal2
 
         windows - python37:
           image.name: 'windows-2019'
           python.version: '3.7'
+          ciwb.build: cp37-win_amd64
+          ciwb.archs: AMD64
         windows - python38:
           image.name: 'windows-2019'
           python.version: '3.8'
+          ciwb.build: cp38-win_amd64
+          ciwb.archs: AMD64
         windows - python39:
           image.name: 'windows-2019'
           python.version: '3.9'
+          ciwb.build: cp39-win_amd64
+          ciwb.archs: AMD64
         windows - python310:
           image.name: 'windows-2019'
           python.version: '3.10'
+          ciwb.build: cp310-win_amd64
+          ciwb.archs: AMD64
 
     steps:
       - task: UsePythonVersion@0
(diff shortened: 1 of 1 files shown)


azure-pipelines-release.yml
```diff
@@ -21,28 +21,44 @@ jobs:
...
         osx - python39:
           image.name: 'macos-12'
           python.version: '3.9'
+          ciwb.build: cp39-macosx_universal2
+          ciwb.archs: universal2
         osx - python310:
           image.name: 'macos-12'
           python.version: '3.10'
+          ciwb.build: cp310-macosx_universal2
+          ciwb.archs: universal2
```
pavlin-policar, line 40: My understanding is that `universal2` packages both ARM and x86 builds into a single wheel, distributing both, making it potentially twice a big (in terms of MB) as individual builds. Wouldn't it make more sense to build individual wheels for both ARM and x86 chipsets? I'm thinking from a pure distribution standpoint here for Orange. The Orange distribution sizes have been a bit big for a while, and wouldn't this just exacerbate the problem?

azure-pipelines-release.yml
```diff
@@ -21,28 +21,44 @@ jobs:
...
 
         windows - python37:
           image.name: 'windows-2019'
           python.version: '3.7'
+          ciwb.build: cp37-win_amd64
+          ciwb.archs: AMD64
```
pavlin-policar, line 46: Won't this make it so that Windows will only have AMD64 builds? I've never worked with `cibuildwheel` before, but the documentation seems to indicate that specifying this flag will only build it for AMD64. I think having a x86 build would be pretty important?

azure-pipelines-release.yml
```diff
@@ -54,11 +70,14 @@ jobs:
       - script: env
         displayName: 'List enviromental variables'
 
-      - script: python -m pip install build
+      - script: python -m pip install cibuildwheel
         displayName: Installing build tools
 
-      - script: python -m build -w .
+      - script: python -m cibuildwheel --output-dir dist/
```
pavlin-policar, line 76: From the documentation it seems like `cibuildwheel` also installs the wheels and tests them afterwards. I have jobs further down that do this exact same thing, but they install optional dependencies first. This would effectively mean testing the wheels twice, no? Is there a way to get `cibuildwheel` to do all the testing by itself, with the optional dependencies?

pavlin-policar (commented):
I've never dealt with `cibuildwheel`, so I have a few questions.

azure-pipelines-release.yml
pavlin-policar: Also, why is `universal2` used only in Python versions 3.9 and 3.10? Why not 3.8 and 3.7 as well? Does `universal2` support only these versions of python?

ales-erjavec: > My understanding is that universal2 packages both ARM and x86 builds into a single wheel, distributing both, making it potentially twice a big (in terms of MB) as individual builds. Wouldn't it make more sense to build individual wheels for both ARM and x86 chipsets? I'm thinking from a pure distribution standpoint here for Orange. The Orange distribution sizes have been a bit big for a while, and wouldn't this just exacerbate the problem?

400KB wont make a difference.

> Does universal2 support only these versions of python?

Yes (3.8 has some support but it is broken).

azure-pipelines-release.yml
ales-erjavec: > Won't this make it so that Windows will only have AMD64 builds? 

Yes. But this matches the existing build configuration.

azure-pipelines-release.yml
ales-erjavec: > From the documentation it seems like cibuildwheel also installs the wheels and tests them afterwards

Only when CIBW_TEST_COMMAND is defined.

PR #229, opened 2023-01-28
Store the last update between optimize() calls

### LINKED ISSUE #228: Bug: running optimize() multiple times produces different result compared to running it once
Running `optimize(n_iter=1)` for 100 times should give the same result as `optimize(n_iter=100)`, but it doesn't. Here is a reproducible example:

```Python
n = 100
np.random.seed(42)
X = np.random.randn(n, 2)

from openTSNE.affinity import PerplexityBasedNN
from openTSNE import TSNEEmbedding
from openTSNE.initialization import random as random_init

A = PerplexityBasedNN(X)
I = random_init(n, random_state=42)

E1 = TSNEEmbedding(I, A, random_state=42)
E2 = TSNEEmbedding(I, A, random_state=42)

E1.optimize(n_iter=100, inplace=True)

for i in range(100):
    E2.optimize(n_iter=1, inplace=True)
    
plt.figure(figsize=(4, 4), layout='constrained')
plt.scatter(E1[:,0], E1[:,1])
plt.scatter(E2[:,0], E2[:,1])
```

tmp-tsne-bug

Maybe it has something to do with how the gains are saved in between the `optimize()` calls?

dkobak:
Fix #228 and add a corresponding test.

### DIFF
--- a/openTSNE/tsne.py
+++ b/openTSNE/tsne.py
@@ -1745,7 +1745,7 @@ class gradient_descent:
                 upper_limit = reference_embedding.box_x_lower_bounds[-1]
 
         if self.update is None:
-            self.update = np.zeros_like(embedding)
+            self.update = np.zeros_like(embedding).view(np.ndarray)
         if self.gains is None:
             self.gains = np.ones_like(embedding).view(np.ndarray)
 
--- a/openTSNE/tsne.py
+++ b/openTSNE/tsne.py
@@ -1829,6 +1829,7 @@ class gradient_descent:
             self.gains[grad_direction_same] = (
                 self.gains[grad_direction_same] * 0.8 + min_gain
             )
+            gradient = gradient.view(np.ndarray)
             self.update = momentum * self.update - learning_rate * self.gains * gradient
 
             # Clip the update sizes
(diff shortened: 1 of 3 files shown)


dkobak:
I don't understand why the pickling test fails here.

pavlin-policar:
The tests likely fail because the pickling/unpickling process is quite brittle. I'll take a look. Thanks for tracking this down. This is definitely something we want to implement.

dkobak:
When the test fails, it happens in `pickle.dumps(embedding)`, and the error is triggered in `__reduce__()`:

```
    957         new_state = state[2] + (
--> 958             self.optimizer,
    959             self.affinities,
    960             self.gradient_descent_params,

AttributeError: 'TSNEEmbedding' object has no attribute 'optimizer'
```

What is weird, is that `embedding.optimizer` actually does exist, and moreover, calling `embedding.__reduce__()` also works fine, but calling `pickle.dumps(embedding)` triggers the above error...

dkobak:
I FOUND THE BUG!!! Only took me two weeks :-) The `update` array needed to be converted from TSNEEmbedding to np.array. All checks pass now.

PR #235, opened 2023-02-15
Fix BH collapse

### LINKED ISSUE #223: Switching spectral initialization to sklean.manifold.SpectralEmbeddings

pavlin-policar:
##### Issue
Fixes #223.

The `is_duplicate` function in the BH implementation had a threshold set to 1e-6, likely taken from other implementations, perhaps scikit-learn. This meant that during the EE phase, where everything was squished together, this condition was triggered a lot, meaning BH wasn't splitting up the points into different quadrants, but was pretending it was dealing with a single point. This mean that there were no repulsive forces being applied to the points at all, leading to an increasing collapse of all the points.

##### Description of changes
I simply changed the threshold for duplicate detection to machine precision, and the issue goes away.
I also copied over the test by @dkobak from #234, but I've also let it run the standard phase. We want to make sure that the embedding isn't collapsed at the end of the optimization, and we don't really care if it is super compressed at the end of the EE phase.

**EDIT**: Not machine precision, because then, duplicate detection actually doesn't work, and the tests fail on iris. Setting the threshold to 1e-16 resolves both issues.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/openTSNE/quad_tree.pyx
+++ b/openTSNE/quad_tree.pyx
@@ -124,7 +124,7 @@ cdef void split_node(Node * node):
     PyMem_Free(new_center)
 
 
-cdef inline bint is_duplicate(Node * node, double * point, double duplicate_eps=1e-6) nogil:
+cdef inline bint is_duplicate(Node * node, double * point, double duplicate_eps=EPSILON) nogil:
     cdef Py_ssize_t d
     for d in range(node.n_dims):
         if fabs(node.center_of_mass[d] - point[d]) >= duplicate_eps:
--- a/openTSNE/quad_tree.pyx
+++ b/openTSNE/quad_tree.pyx
@@ -124,7 +124,7 @@ cdef void split_node(Node * node):
     PyMem_Free(new_center)
 
 
-cdef inline bint is_duplicate(Node * node, double * point, double duplicate_eps=EPSILON) nogil:
+cdef inline bint is_duplicate(Node * node, double * point, double duplicate_eps=1e-16) nogil:
     cdef Py_ssize_t d
     for d in range(node.n_dims):
         if fabs(node.center_of_mass[d] - point[d]) >= duplicate_eps:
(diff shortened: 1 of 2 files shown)


dkobak:
With all the recent changes it would be great to release a new version (0.6.3 or 0.7), what do you think? Some of the changes (like new learning rate defaults) affect the outcome with default params.

pavlin-policar:
Yes, there have been a few changes since 0.6.2. We've added jittering which actually does change the resulting embeddings a bit, so this probably warrants an 0.7.0 release. The rest was mainly bugfixes I think.

I'd like to get #226 solved first, and there are some issues on conda, but once these get resolved, I'll make a new release.

dkobak:
Yes, init jittering, but also learning rate that is now set to n/exag in each epoch, and momentum fixed to 0.8 throughout.

Cool.

PR #242, opened 2023-05-12
Implement binary uniform affinity

dkobak:
This is a very minor suggestion, but I realized some time ago that in some papers we implement "uniform affinities" slightly differently: not as `(P + P.T) / 2` but as `(P + P.T > 0).astype(float)` where `P` is the kNN adjacency matrix. The only difference is that the former has two possible non-zero values, whereas the latter has only one possible non-zero value, which is kind of neat.

This does not really make much of a different in practice, but I thought it may be nice to implement it in `Uniform()` using `symmetrize="average"` and `symmetrize="or"`, in addition to `symmetrize=True/False` which this class has anyway.

tsne-uniform-two-kinds

### DIFF
--- a/openTSNE/affinity.py
+++ b/openTSNE/affinity.py
@@ -1197,6 +1197,12 @@ class Uniform(Affinities):
             P = (P + P.T > 0).astype(float)
         elif symmetrize == "mean":
             P = (P + P.T) / 2
+        elif symmetrize == False:
+            pass
+        else:
+            raise ValueError(
+                f"Symmetrization method ({symmetrize}) is not recognized."
+            )
 
         # Convert weights to probabilities
         P /= np.sum(P)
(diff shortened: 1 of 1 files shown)


pavlin-policar:
I like this, but perhaps instead of "average" and "or", I would prefer "mean", and "max", which (to me) feel more standard. If we do go this route, I would also change the default parameter value from `symmetrize=True` to `symmetrize="max"`. Do you think it would make more sense to default to the binary affinity kernel, or the current "mean" symmetrization?

pavlin-policar:
I hate to be nit-picky, but could you please add a case if we get a string that isn't either "mean" or "max" and raise a `ValueError` with something like `symmetrization {method} is not recognized`. Just so we don't fail silently in case the user makes a typo or something like that.

dkobak:
Makes sense. I changed it so that the default is True (that's how it was before) and this behaves like "mean" but triggers a FutureWarning. In the future one can remove the boolean parameters altogether, and only keep "mean"/"max"/"none".

PR #265, opened 2024-11-20
Add knn_kwargs parameter to TSNE API

### LINKED ISSUE #256: Allow KNNIndex classes to take **kwargs arguments and pass them to KNN library
See title. We discussed it at Dagstuhl. Should be useful for PyNNdescent, but could potentially be useful for Annoy/HNSW/etc at well.

We could potentially allow Affinity classes to take **kwargs as well and pass them to the KNNIndex. Here I'm less sure if that's a good design idea.

jnboehm:
##### Issue

Discussed this with @dkobak earlier today and saw that there is already an issue present. 

This PR fixes #256 

##### Description of changes

We add `knn_kwargs` as a dict and pass that through to where the knn index is created.  I added some parameters to Annoy (`n_trees`) and Pynndescent (`n_trees`, `n_iter`, `max_candidates`).

##### Includes
- [X] Code changes
- [ ] Tests
- [X] Documentation (docstring)

### DIFF
--- a/openTSNE/nearest_neighbors.py
+++ b/openTSNE/nearest_neighbors.py
@@ -222,7 +222,7 @@ class Annoy(KNNIndex):
     def __init__(self, *args, knn_kwargs=None, **kwargs):
         super().__init__(*args, **kwargs)
         self.knn_kwargs = dict() if knn_kwargs is None else knn_kwargs.copy()
-        self.n_trees = knn_kwargs.pop("n_trees", 50)
+        self.n_trees = self.knn_kwargs.pop("n_trees", 50)
 
     def build(self):
         data, k = self.data, self.k


openTSNE/affinity.py
```diff
@@ -777,6 +795,7 @@ def __init__(
         n_jobs=1,
         random_state=None,
         verbose=False,
+        knn_kwargs=None,
```
pavlin-policar, line 798: From what I can tell, you've added the `knn_kwargs` parameter to the `PerplexityBasedAffinities` and `MultiscaleMixture` classes. Could you please add them to the other Affinity types as well? This would certainly come in useful, e.g., in the `UniformAffinities` class. And I'd want it in the other affinity classes as well for consistency.

openTSNE/nearest_neighbors.py
```diff
@@ -470,8 +476,14 @@ def build(self):
         timer.__enter__()
 
         # These values were taken from UMAP, which we assume to be sensible defaults
-        n_trees = 5 + int(round((data.shape[0]) ** 0.5 / 20))
-        n_iters = max(5, int(round(np.log2(data.shape[0]))))
+        if self.n_trees is None:
```
pavlin-policar, line 479: So this bit of code here looks like it would only support three parameters for NNDescent: `n_trees`, `n_iters`, and `max_candidates`. pynndescent supports a whole range of other parameters as well:
https://github.com/lmcinnes/pynndescent/blob/master/pynndescent/pynndescent_.py#L673-L697

Ideally, I'd be able to specify any of these arguments. Instead of storing the nndescent-specific parameters as class attributes (`self.n_trees`, `self.n_iters`), I think it would be better to store a dictionary of `self.knn_kwargs`, and use that. In this case, in case a KNN library deprecates its parameters and stops using it, we won't have to go through a deprecation cycle as well. And this would make specifying any potential newly introduced arguments as simple as just passing it into the dictionary without having to update openTSNE. I like this approach better since openTSNE remains mostly agnostic to the KNN libraries, which can update their parameters independently of openTSNE.

So maybe running something like `NNDescent(data, k=15, knn_kwargs={"n_trees": 50})` instead of the way you have it now: NNDescent(data, k=15, n_trees=50).

As a side benefit, this would also make it explicit to users that this is a dict that gets passed through to the underlying implementation, and that we don't handle the parameters in any special way. Similarly to the logic of `metric_params`.

openTSNE/nearest_neighbors.py
```diff
@@ -417,7 +418,9 @@ class NNDescent(KNNIndex):
         "yule",
     ]
 
-    def __init__(self, *args, **kwargs):
+    def __init__(
```
pavlin-policar, line 421: Similarly to the Affinity classes, we'd want the option to pass in `knn_kwargs` to any KNN index. Here, I think you've missed the HNSW index and scikit-learn index.

openTSNE/nearest_neighbors.py
```diff
@@ -217,8 +217,9 @@ class Annoy(KNNIndex):
         "taxicab",
     ]
 
-    def __init__(self, *args, **kwargs):
+    def __init__(self, *args, n_trees=50, **kwargs):
```
pavlin-policar, line 220: See the NNDescent comment below. Ideally, this would come through a `knn_kwargs` parameter containing a dictionary of parameters, so as to avoid having to keep these classes in sync with the underlying Annoy implementation.

pavlin-policar (changes requested):
Hi Jan, thanks for the PR! I've been meaning to get around to this for quite some time, so I really appreciate you taking the time to take care of this.

I do have a few comments though before merging.

pavlin-policar:
The failing windows tests likely have nothing to do with your changes, it's probably just standard CI problems that will go away in future runs...

openTSNE/nearest_neighbors.py
jnboehm: Okay, that sounds like a good idea to me.  I took that approach so that we can still use the `**kwargs` to pass parameters to the super class.  I like the approach with preserving the dict instead and only splatting it later down the line for all classes.  

Should the default parameters be preserved?  As in, should we check if they are present in the dict and if not take the defaults that are currently used?  I personally would like to use the default of pynndescent, but that would potentially break backwards compatibility, so I would tend to do the check and by default initialize with the current defaults still.

jnboehm:
I'm working on it now, but I noticed that the classes also have a `__setstate__` method (for pickling if I am not mistaken?).  Unfortunately, I do not really know how that would interact with the `knn_kwargs`, or is that unrelated?

openTSNE/nearest_neighbors.py
jnboehm: Passed is to the `Index` class for HNSW and to `NearestNeighbors` for sklearn

openTSNE/nearest_neighbors.py
jnboehm: I have now reworked this a bit.  Not sure if the current state is satisfying because we now pass in `n_trees` as an entry in the `knn_kwargs` dict, but the value actually goes into `index.build(...)`.

openTSNE/affinity.py
jnboehm: Oh, I missed that, have updated the code.

jnboehm:
Should I also add some tests for this?  I'm not sure how exactly they should look, but I'd guess that they should create a `TSNE` object with a nonempty `knn_kwargs` parameter.  Would this already check if it can be properly pickled through `__setstate__` and related functionality?  To be honest, the test logic looks a bit mysterious to me and I am not really familiar with it, unfortunately.

pavlin-policar:
Hey Jan, yes, some tests would definitely be nice to check that things are working as intended. That would probably involve patching the NN indices and making sure they were, indeed, called with the correct `knn_kwargs` parameter, then going through all possible ways to actually call this method.

Looking at the [code](https://github.com/pavlin-policar/openTSNE/blob/master/openTSNE/nearest_neighbors.py#L315-L332), pickling should work out of the box, since all class attributes should be saved [here](https://github.com/pavlin-policar/openTSNE/blob/master/openTSNE/nearest_neighbors.py#L320). Ideally, you'd add a test that checks this too. Looking at the tests, however, I wasn't very good at adding tests for pickling. I would add it to the [`KNNIndexTestMixin`](https://github.com/pavlin-policar/openTSNE/blob/master/tests/test_nearest_neighbors.py#L20) class, so that it runs for all the different NN index types.

PR #271, opened 2025-04-07
Features/lmn/290325 affinity

luanamarinho:
##### Changes
- Refactored MultiscaleMixture:
- Vectorize perplexity validation using NumPy
- Enhanced validation with edge case handling

##### Tests
- Added test for MultiscaleMixture.check_perplexities


pavlin-policar:
Thanks for the PR. However, I'm not entirely sure what problem this is solving. Could you please provide a paragraph or so explaining the need for these changes and what problems they solve.

From the code, it seems like there are a whole lot of minor changes. The biggest change seems to be that duplicates are now removed from perplexity lists and they are sorted, is that right? Some of these changes are certainly welcome, but I'm not entirely sure all are needed.

Could you also please take a look at the failing tests on Windows? The error `Windows fatal exception: access violation` is cryptic, but it appears to consistently fail at the added test, so there must be something there that Winsdows doesn't like.

luanamarinho:
Hi Pavlin. Thanks for reviewing the PR so promptly. You' right to note these changes focus on broader quality improvements, rather than targeting a specific bug. The commits mean to make the class more robust, while maintaining all existing functionality. Here's a summary of the key modifications:

1. Validation centralization
     **Before**: Perplexity validation was repetitive and scattered across `__init__`, `set_perplexities`, and `to_new`, with redundant `min(n_samples - 1, int(3 * max_perplexity)`, violating SRP and DRY principles.
     **Now**: check_perplexities is the single authoritative source for:
                   Input type checking (np.ndarray conversion);
                   Dimensionality handling (.ravel() for scalars);
                   Value validation (positivity, max bounds) - including that of knn.
     **Impact:** Changes to validation logic now require edits in only one place.
3. Robust perplexity management in `set_perplexities`
    **Fixed** flawed deduplication (silent bug);
     Moved bounds validation entirely into `check_perplexities` (previously done ad-hoc): SRP and DRY principles;
     Added early validation to **fail fast** - potential RuntimeErrors are now caught during initial checks instead of mid-execution.
5. Optimized `check_perplexities` implementation
Replaced manual loops with vectorized NumPy operations - improved efficiency;
Simplified logic while adding proper edge case handling - **fail fast**.
7. Defensive Programming
    Wrapped all critical operations in structured error handling for better debuggability
    Ensured validation occurs before expensive computation begins - fail fast

These changes eliminate subtle maintenance hazards while improving efficiency. I’m currently verifying the Windows-specific issue separately and will update the PR accordingly. Happy to discuss or adjust any aspect!

Kind regards,
Luana

PR #275, opened 2025-10-22
Add learnable dof

Artemii-Shlychkov:
##### Issue
T-distributed Stochastic Neighbor Embedding (t-SNE) is a widely used tool for dimensionality reduction and visualization of high-dimensional datasets. By replacing the Gaussian kernel in SNE with a Cauchy kernel (Student t-distribution with the degree-of-freedom (dof) parameter alpha set to 1), it alleviates the “crowding problem” in low-dimensional embeddings. Varying the degree-of-freedom parameter alpha affects t-SNE embeddings on both toy and real-world datasets (e.g., MNIST and single-cell RNA sequencing). Moreover, alpha can be regarded as a trainable parameter, allowing it to be adjusted during embedding optimization via a gradient-based method. Alpha values different from 1 can yield superior embeddings, reflected by reduced Kullback-Leibler (KL) divergence and higher k-Nearest Neighbors (kNN) recall scores, at least on some datasets. Overall, this suggests that alpha optimization can lead to more faithful low-dimensional representations of high-dimensional data.

##### Description of changes
 `_tsne.pyx`
Main changes with regards to dof optimization:
- `estimate_positive_gradient_nn` function:
    - added computations for the alpha gradient positive term
- `_estimate_negative_gradient_single` function
    - added computations for the alpha gradient negative term
- `estimate_negative_gradient_bh` function
    - added normalization of alpha gradient negative term on `sum_Q`

 `tsne.py`
- added dataclass `OptimizationStats` to track changes in KL-divergence, dof values, alpha gradient values and embeddings with every iteration
- `kl_divergence_bh` function
    - added dof gradient computations (`alpha_grad`), based on the ouputs of _tsne module
- `gradient_descent` class
    - added optional `optimize_for_alpha` bool argument to trigger dof-optimization
    - added optional `dof_lr` argument 
    - added optional `dof` update with current value of `alpha_grad` and `dof_lr` learning rate
    - `eval_error_every_iter` argument to make tracking of KL-divergence more flexible
##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


openTSNE/tsne.py
```diff
@@ -1593,6 +1616,8 @@ def __call__(
         momentum=0.8,
         exaggeration=None,
         dof=1,
+        optimize_for_alpha=False,
+        dof_lr=0.5,
```
pavlin-policar: Can we connect this to the t-SNE learning rate somehow? Don't like the extra parameter

openTSNE/tsne.py
```diff
@@ -1593,6 +1616,8 @@ def __call__(
         momentum=0.8,
         exaggeration=None,
         dof=1,
+        optimize_for_alpha=False,
```
pavlin-policar: activate with dof="auto" for optimization

openTSNE/tsne.py
```diff
@@ -1798,6 +1841,18 @@ def __call__(
                 should_eval_error=should_eval_error,
             )
 
+            if optimize_for_alpha:
+                dof -= dof_lr * alpha_grad
```
pavlin-policar: This is a new optimizer. Can we optimize with the existing delta-bar-delta optimizer? Let's look at some loss curves.

pavlin-policar:
**Added**

- dof="auto" learns dof during optimization; dof_ records the value used; initial_dof sets the warm-start
- add `transform` functionality
- change kl_divergence to sklearn-style attribute; optimization_iters_ (cumulative iteration counter); IterationState callback snapshots
- "Degrees of freedom" guide page + swiss-roll example notebook

**Transform**
- Three-mode dof on transform/prepare_partial: "inherit" (default), "auto", or a number; the dof mode carries over across optimize() calls
- transform(dof="auto") diverged to large negative values: the dof-gradient's negative term used global rather than per-point normalization
- transform(dof="auto") drifts upwards. Clearly unstable optimization. Emits a warning; "inherit" recommended

**Changed**

- Extracted DeltaBarDeltaOptimizer; dof gradients computed only when learning
- Changed callback signature
- Versioned dict-based embedding pickling with legacy migration

**CI / tooling**
- Min Python 3.9 → 3.10; CI installs via extras; docs pipeline palette-quantizes example PNGs.

**Still TODO**: Remove the learnable_dof example notebook.

docs/source/degrees_of_freedom.rst
```diff
@@ -0,0 +1,352 @@
...
+   * - ``"inherit"`` *(default)*
+     - Use the parent's ``dof_``, held fixed.
+   * - ``"auto"``
+     - Learn a new value, warm-started from the parent.
+   * - ``"auto"`` with ``initial_dof``
+     - Learn a new value, warm-started from ``initial_dof``.
+   * - a number
+     - Hold the degrees of freedom fixed at that number.
+
+A worked example: the swiss roll
```
dkobak: I think "Swiss" should be capitalized. Also several times it is used below.

docs/source/degrees_of_freedom.rst
```diff
@@ -0,0 +1,352 @@
...
+.. image:: examples/07_degrees_of_freedom/output_17_0.png
+   :align: center
+
+Building the embedding by hand
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The estimator above is a convenient wrapper. For finer control the affinities,
+the initialization, and the optimizer can be assembled directly and optimized in
+phases. The ``dof`` mode can even be varied between phases. Here we hold
+``dof=1`` (standard t-SNE) during early exaggeration, while the broad layout of
```
dkobak, line 269: This sentence sounds as if everything shown above did not use this approach (i.e. dof was also learned during the early exaggeration phase in the above Swiss roll experiments). But in Artemii's initial implementation, dof was frozen during early exaggeration and only learned after the early exaggeration was turned off. Is it not the case after your edits, Pavlin? I am not saying what is better or worse here, just clarifying.

docs/source/degrees_of_freedom.rst
```diff
@@ -0,0 +1,352 @@
...
+reference it tends to a larger value here (around ``19``), while a fixed
+``dof=1`` deliberately uses a heavier-tailed kernel than the reference.
+
+.. code-block:: python
+
+    new_inherit = reference.transform(x_new)               # 5.4
+    new_auto    = reference.transform(x_new, dof="auto")   # ~19, drifts upward
+    new_fixed   = reference.transform(x_new, dof=1)        # 1.0
+
+.. image:: examples/07_degrees_of_freedom/output_28_0.png
```
dkobak, line 345: The embeddings appear identical on this image. Are they actually any different, or is it within machine precision? I have not thought it through, I am not sure what `dof` does here, when the location of each test point is optimized separately.

docs/source/degrees_of_freedom.rst
```diff
@@ -0,0 +1,352 @@
...
+.. image:: examples/07_degrees_of_freedom/output_25_0.png
+   :align: center
+   :width: 60%
+
+The new points (outlined) slot smoothly into the reference (drawn faintly),
+continuing its colour gradient. The other modes behave as described above:
+``dof="auto"`` relearns a value for the new points, and a number pins them to a
+fixed value. Because ``"auto"`` is only weakly identified against a fixed
+reference it tends to a larger value here (around ``19``), while a fixed
+``dof=1`` deliberately uses a heavier-tailed kernel than the reference.
```
dkobak: I would explicitly add a sentence saying that the embedding itself is barely affected.

dkobak (commented):
Some comments on the new docs page.

tests/test_dof_learning.py
```diff
@@ -0,0 +1,986 @@
...
+            warnings.simplefilter("always")
+            value = emb.kl_divergence
+        self.assertTrue(
+            any(issubclass(w.category, DeprecationWarning) for w in caught),
+            "Reading `kl_divergence` should emit a DeprecationWarning.",
+        )
+        self.assertEqual(value, emb.kl_divergence_)
+
+
+class TestDofAutoLearning(unittest.TestCase):
```
dkobak, line 204: An impressive set of tests! But is there a test that checks that the final KL divergence (on some small dataset) with dof="auto" is smaller than with dof=1? I think that would be good to have.

dkobak (commented):
Looked at the tests.

docs/source/degrees_of_freedom.rst
```diff
@@ -0,0 +1,352 @@
...
+    standard = TSNE(dof=1, random_state=42).fit(x)
+    learned  = TSNE(dof="auto", random_state=42).fit(x)
+
+    learned.dof_   # ~5.0
+
+Side by side, the three embeddings show the progression from a fragmented
+heavy-tailed embedding, through standard t-SNE, to the smooth unrolled sheet
+recovered with the learned degrees of freedom.
+
+.. image:: examples/07_degrees_of_freedom/output_17_0.png
```
dkobak, line 260: It would be great to add somewhere here a result for `TSNE(dof="auto", initialization="spectral")` if it does produce an embedding without "twists", as I would hope it does.

dkobak (commented):
One more suggestion for the degrees_of_freedom notebook and rst file.

dkobak:
Just saw the spectral-init result -- wow, I did not expect that the estimated dof will get so much higher in this case! This is actually a great result, fits perfectly to our story.

docs/source/degrees_of_freedom.rst
pavlin-policar: This should be fixed everywhere now.

docs/source/degrees_of_freedom.rst
pavlin-policar: You're right -- that is how Artemii implemented it. I missed that. Yes, in the current implementation, dof="auto" learns the dof through both phases of the optimization. I've now run a whole host of experiments to make sure that the optimization is stable at any exaggeration (which was not easy!) and dof="auto" learns through all phases. I think this is crucial because we can run t-SNE with exaggeration=12 in EE and then exaggeration=4 in the regular phase. So the dof learning needs to be stable even under that regime.

tests/test_dof_learning.py
pavlin-policar: We can thank claude for that :) I added this test later on.

docs/source/degrees_of_freedom.rst
pavlin-policar: I've added it at the end. Interestingly, it also sort of depends on the spectral init random seed. For instance, setting the random seed to 42 still resulted in a twisted embedding, while setting it to 0 removed that twist.

docs/source/degrees_of_freedom.rst
pavlin-policar: Added this.

openTSNE/tsne.py
dkobak: this review comment is obsolete and can be resolved

docs/source/degrees_of_freedom.rst
dkobak: Is this comment not obsolete? Did you still want to look into this @Pavlin? If the embeddings are not identical, then it's fine, can be resolved. I just wanted to double-check.

PR #281, opened 2025-11-03
Switch to tempfile.NamedTemporaryFile

### LINKED ISSUE #210: [Windows] save TSNEEmbedding to binary, Directory error
##### Expected behaviour
On Windows OS, when trying to save the TSNEEmbedding  object, or affinities, tried to save it with 
`pickle.dump(embeddings,open(os.path.join(self.models_path,"tsne_global_embeddings.sav"),"wb"))`
or also tried to save as array to reconstruct the object later using `numpy.save("file.npy",affinities)`

These lines both work just fine under linux distributions what I tried.
But loading them back on Windows breaks with the same error as on the save methods, both scenario. 

##### Actual behaviour
Windows OS can't find/create the temporary directory/files when trying to touch the file. unfortunately I haven't had more time to look deeper into it yet, what could cause this behaviour.

`*** NotADirectoryError: [WinError 267] The directory name is invalid: 'C:\\Users\\tomcs\\AppData\\Local\\Temp\\tmp7biujwdz\\tmp.ann`
##### Steps to reproduce the behavior
opentsne==0.6.2

I think this would be the same with most settings. although I am using the following settings to train before trying to save.

`affinities = openTSNE.affinity.PerplexityBasedNN(
            X,
            perplexity=500,
            n_jobs=32,
            random_state=0,
        )`
`init = openTSNE.initialization.pca(X,n_components=3, random_state=42)`
`tsne = openTSNE.TSNE(3,
            exaggeration=None,
            n_jobs=16,
            verbose=True,
            negative_gradient_method ="bh"
        )`
        `embeddings = tsne.fit(affinities=affinities, initialization=init)`
`pickle.dump(embeddings,open("tsne_global_embeddings.sav","wb"))`

dkobak:
Trying to fix #210 as discussed there.

See https://docs.python.org/3.12/library/tempfile.html#examples for an official example of how to work with `NamedTemporaryFile` when one needs to write smth into it and then read from it:

```Python
# create a temporary file using a context manager
# close the file, use the name to open the file again

with tempfile.NamedTemporaryFile(delete_on_close=False) as fp:
    fp.write(b'Hello world!')
    fp.close()
    # the file is closed, but not removed
    
    # open the file again by using its name
    with open(fp.name, mode='rb') as f:
        f.read()
# file is now removed

### DIFF
--- a/openTSNE/nearest_neighbors.py
+++ b/openTSNE/nearest_neighbors.py
@@ -338,7 +338,7 @@ class Annoy(KNNIndex):
         if self.index is not None:
             with tempfile.NamedTemporaryFile(delete_on_close=False) as tmp:
                 self.index.save(tmp.name)
-                # tmp.close() # Just trying out if maybe it's closed already
+                # tmp.close() # Windows test fails independent of including/excluding this line
 
                 with open(tmp.name, "rb") as f:
                     b64_index = base64.b64encode(f.read())
@@ -373,7 +373,6 @@ class Annoy(KNNIndex):
 
             self.index = AnnoyIndex(state["data"].shape[1], annoy_metric)
             with tempfile.NamedTemporaryFile(delete_on_close=False) as tmp:
-                # with open(tmp.name, "wb") as f:
                 tmp.write(base64.b64decode(b64_index))
                 tmp.close()
                 
@@ -709,7 +708,6 @@ class HNSW(KNNIndex):
 
             self.index = Index(space=hnsw_metric, dim=state["data"].data.shape[1])
             with tempfile.NamedTemporaryFile(delete_on_close=False) as tmp:
-                # with open(tmp.name, "wb") as f:
                 tmp.write(base64.b64decode(b64_index))
                 tmp.close()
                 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by dkobak** [comment]
The current state of this PR is that the tests pass on Linux and OSX (for Python >= 3.12) but fail on Windows. Curiously, only one single test fails:

```
tests/test_nearest_neighbors.py::TestAnnoy::test_pickle_with_built_index FAILED [ 41%]
tests/test_nearest_neighbors.py::TestAnnoy::test_pickle_without_built_index PASSED [ 41%]
tests/test_nearest_neighbors.py::TestAnnoy::test_pickle_without_built_index_cleans_up_fname PASSED [ 42%]
tests/test_nearest_neighbors.py::TestHNSW::test_pickle_with_built_index PASSED  [ 52%]
tests/test_nearest_neighbors.py::TestHNSW::test_pickle_without_built_index PASSED [ 53%]
tests/test_nearest_neighbors.py::TestHNSW::test_pickle_without_built_index_cleans_up_fname PASSED [ 54%]
```

Note that `test_pickle_with_built_index` passes for HNSW but fails for Annoy! Even though the tests are identical, and `__getstate__()` is implemented the same way for HNSW and Annoy.

pavlin-policar:
So, as we discussed, the state of this PR is as follows:
- this makes a step in the right direction. It fixes the base problem, but some tests still fail for some reason
- the fix includes a Python >=3.12 feature while openTSNE currently still supports 3.10. The end-of-life of Python 3.11 is sometime at the end of 2027, so this is not an entirely viable fix until then. We can wrap in an if statement or something, but it's not clean

So, from my understanding, this is a step in the right direction, but not something that can be merged.

PR #284, opened 2026-03-17
Fix docstring and error messages

dkobak:
Small docstring and error message fixes in affinity classes.


dkobak:
@pavlin-policar Some tests are failing here, but I don't think it's due to the changes I made in this PR...

